PR #1, opened 2024-06-14
Add `package:dart_model` with Json and JsonChanges classes.

davidmorgan:
Add some first `package:data_model`.

I wanted to avoid anything with schema / generation in the first checkin, so this is just some general JSON handling. The `JsonChanges` extension types probably _will_ end up in a schema, but I wanted to start somewhere :)

### DIFF
--- a/pkgs/dart_model/lib/src/json.dart
+++ b/pkgs/dart_model/lib/src/json.dart
@@ -12,51 +12,43 @@ import 'json_changes.dart';
 ///
 /// Views onto the data are provided as [Map] instances, see [asMap].
 final class JsonData {
-  Map<String, Object?> _root = {};
+  final Map<String, Object?> _root;
 
   /// Instantiates with a deep copy of [json].
   ///
   /// Throws if any values not allowed in JSON data are present.
   ///
   /// TODO(davidmorgan): add a way to build without copying.
-  JsonData.deepCopyAndCheck(Map<String, Object?> json) {
-    _deepCopyAndCheckMap(json, _root);
-  }
+  JsonData.deepCopyAndCheck(Map<String, Object?> json)
+      : _root = _deepCopyAndCheckMap(json);
 
   /// A `Map` view onto the root of this JSON data.
   Map<String, Object?> get asMap => JsonMap._(_root);
 
-  static void _deepCopyAndCheckMap(
-      Map<String, Object?> from, Map<String, Object?> to) {
-    from.forEach((key, value) {
-      to[key] = _deepCopyAndCheckValue(value);
+  static Map<String, Object?> _deepCopyAndCheckMap(Map<String, Object?> map) {
+    final result = <String, Object?>{};
+    map.forEach((key, value) {
+      result[key] = _deepCopyAndCheckValue(value);
     });
+    return result;
   }
 
-  static void _deepCopyAndCheckList(List<Object?> from, List<Object?> to) {
-    for (final value in from) {
-      to.add(_deepCopyAndCheckValue(value));
+  static List<Object?> _deepCopyAndCheckList(List<Object?> list) {
+    final result = <Object?>[];
+    for (final value in list) {
+      result.add(_deepCopyAndCheckValue(value));
     }
+    return result;
   }
 
   static Object? _deepCopyAndCheckValue(Object? value) {
-    if (value is Map<String, Object?>) {
-      final copy = <String, Object?>{};
-      _deepCopyAndCheckMap(value, copy);
-      return copy;
-    } else if (value is String ||
-        value is bool ||
-        value is num ||
-        value == null) {
-      return value;
-    } else if (value is List<Object?>) {
-      final copy = <Object?>[];
-      _deepCopyAndCheckList(value, copy);
-      return copy;
-    } else {
-      throw UnsupportedError(
-          'JsonData cannot hold value of type ${value.runtimeType}: $value');
-    }
+    return switch (value) {
+      Map<String, Object?>() => _deepCopyAndCheckMap(value),
+      List<Object?>() => _deepCopyAndCheckList(value),
+      String() || bool() || num() || null => value,
+      Object() => throw UnsupportedError(
+          'JsonData cannot hold value of type ${value.runtimeType}: $value'),
+    };
   }
 
   /// Computes [JsonChanges] that is this data minus [previous].
@@ -154,12 +146,11 @@ final class JsonData {
       required Map<String, Object?> updates}) {
     updates.forEach((key, value) {
       if (value is Map<String, Object?>) {
-        if (root[key] is! Map<String, Object?>) {
-          root[key] = <String, Object?>{};
+        var node = root[key];
+        if (node is! Map<String, Object?>) {
+          root[key] = node = <String, Object?>{};
         }
-        _applyUpdates(
-            root: (root[key] ??= <String, Object?>{}) as Map<String, Object?>,
-            updates: value);
+        _applyUpdates(root: node, updates: value);
       } else {
         root[key] = value;
       }
@@ -184,6 +175,8 @@ final class JsonData {
 }
 
 /// An immutable `Map` view onto part of [JsonData].
+// TODO(davidmorgan): check whether having a custom map implementation
+// here causes any performance regression.
 final class JsonMap
     with MapMixin<String, Object?>
     implements Map<String, Object?> {
--- a/pkgs/dart_model/lib/src/json.dart
+++ b/pkgs/dart_model/lib/src/json.dart
@@ -1,0 +1,113 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+
+import 'json_changes.dart';
+
+/// JSON data.
+final class Json {
+  Map<String, Object?> _root;
+
+  Json.fromJson(Map<String, Object?> json) : _root = json;
+
+  JsonChanges computeChangesFrom(Json previous) {
+    final updates = <Update>[];
+    final removals = <Removal>[];
+    _compute(previous._root, _root, Path([]), updates, removals);
+    return JsonChanges(updates: updates, removals: removals);
+  }
+
+  static void _compute(
+      Map<String, Object?> previous,
+      Map<String, Object?> current,
+      Path path,
+      List<Update> updates,
+      List<Removal> removals) {
+    for (final key in previous.keys.followedBy(current.keys).toSet()) {
+      final keyIsInPrevious = previous.containsKey(key);
+      final keyIsInCurrent = current.containsKey(key);
+
+      if (keyIsInPrevious && !keyIsInCurrent) {
+        removals.add(Removal(path: path.followedByOne(key)));
+      } else if (keyIsInPrevious && keyIsInCurrent) {
+        // It's either the same or a change.
+        final previousValue = previous[key]!;
+        final currentValue = current[key]!;
+
+        if (currentValue is Map<String, Object?>) {
+          if (previousValue is Map<String, Object?>) {
+            _compute(previousValue, currentValue, path.followedByOne(key),
+                updates, removals);
+          } else {
+            updates.add(
+                Update(path: path.followedByOne(key), value: currentValue));
+          }
+        } else if (currentValue is String) {
+          if (previousValue is! String || previousValue != currentValue) {
+            updates.add(
+                Update(path: path.followedByOne(key), value: currentValue));
+          }
+        } else if (currentValue is List) {
+          if (previousValue is! List ||
+              !const DeepCollectionEquality()
+                  .equals(previousValue, currentValue)) {
+            updates.add(
+                Update(path: path.followedByOne(key), value: currentValue));
+          }
+        } else {
+          throw UnsupportedError(
+              'Unsupported change: $previousValue to $currentValue');
+        }
+      } else {
+        // It's new.
+        updates.add(Update(path: path.followedByOne(key), value: current[key]));
+      }
+    }
+  }
+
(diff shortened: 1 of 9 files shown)


pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,113 @@
...
+      Map<String, Object?> current,
+      Path path,
+      List<Update> updates,
+      List<Removal> removals) {
+    for (final key in previous.keys.followedBy(current.keys).toSet()) {
+      final keyIsInPrevious = previous.containsKey(key);
+      final keyIsInCurrent = current.containsKey(key);
+
+      if (keyIsInPrevious && !keyIsInCurrent) {
+        removals.add(Removal(path: path.followedByOne(key)));
```
jakemac53, line 33: Do we plan to have `null` be a meaningful value in the schema for anything? If not, we could merge the Update/Removal classes and just have the value be null instead for removals. Or, even go so far as removing JsonChanges and representing the diff as Json with null entries for removals.

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,113 @@
...
+      } else if (keyIsInPrevious && keyIsInCurrent) {
+        // It's either the same or a change.
+        final previousValue = previous[key]!;
+        final currentValue = current[key]!;
+
+        if (currentValue is Map<String, Object?>) {
```
jakemac53, line 39: does this not need to handle all Json types?

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,113 @@
...
+            updates.add(
+                Update(path: path.followedByOne(key), value: currentValue));
+          }
+        } else if (currentValue is String) {
+          if (previousValue is! String || previousValue != currentValue) {
+            updates.add(
+                Update(path: path.followedByOne(key), value: currentValue));
+          }
+        } else if (currentValue is List) {
+          if (previousValue is! List ||
```
jakemac53, line 53: Don't we probably want incremental updates for lists too?

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+  }
+
+  /// Returns a new [Json] instance with [changes] made.
+  ///
+  /// TODO(davidmorgan): actually return a new model, copying as needd.
+  Json change(JsonChanges changes) {
+    for (final update in changes.updates) {
+      _updateAtPath(_root, update.path, update.value);
```
jakemac53, line 75: This model for updates requires re-traversing the model through the same exact paths a lot.

We should structure the updates as a tree, so we can walk the model once and make updates as we go, similar to how the updates themselves are constructed.

pkgs/dart_model/lib/src/json_changes.dart
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'json.dart';
+
+/// The changes between two [Json] instances.
+extension type JsonChanges.fromJson(Map<String, Object?> node) {
```
jakemac53, line 8: Oof, storing this as an object with keys is physically painful to me 😬

pkgs/dart_model/lib/src/json_changes.dart
```diff
@@ -0,0 +1,58 @@
...
+extension type Path.fromJson(List<Object?> node) {
+  Path(List<String> path) : this.fromJson(path);
+
+  List<String> get path => (node as List).cast();
+
+  String? get uri => path.isEmpty ? null : path.first;
+
+  /// This path followed by [element].
+  Path followedByOne(String element) =>
+      Path(path.followedBy([element]).toList());
```
jakemac53, line 34:
```suggestion
      Path(path.toList()..add(element));
```

pkgs/dart_model/lib/src/json_changes.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  /// This path except the first entry.
+  Path skipOne() => Path(path.skip(1).toList());
```
jakemac53, line 37: If Path was an extension type on iterable then at least this could be much faster and avoid the copy. We could also probably avoid a lot of the other list copying by removing the `toList()` calls. Serializing to JSON would still be slow though with tons of duplicate logic in there.

There is in general way to much copying of lists going on in this class imo and a different approach is needed.

pkgs/dart_model/lib/src/json_changes.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  /// The nodes that were updated by these changes.
+  List<Update> get updates => (node['updates']! as List).cast();
+
+  /// The nodes that were removed by these changes.
+  List<Removal> get removals => (node['removals']! as List).cast();
+}
+
+/// A path into JSON data.
+extension type Path.fromJson(List<Object?> node) {
```
jakemac53, line 25: These paths are extremely expensive. They should be represented as linked lists or something? I am not sure how exactly to translate that into JSON in the best way, but I think it is critical that we don't duplicate the entire path for each update, copying the lists etc.

pkgs/dart_model/lib/src/json_changes.dart
```diff
@@ -0,0 +1,58 @@
...
+  Object? get value => node[1];
+}
+
+/// One removal from JSON data: a [Path].
+extension type Removal.fromJson(List<Object?> node) {
+  Removal({
+    required Path path,
+  }) : this.fromJson(path.path);
+
+  Path get path => node as Path;
```
jakemac53, line 57: In general you shouldn't cast to extension types, just call the constructor (same elsewhere).

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+
+import 'json_changes.dart';
+
+/// JSON data.
```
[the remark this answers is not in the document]
dave26199: This is a crucial part of the design but still a bit fuzzy, let me try writing it out once to see if that helps :)

This class is to encapsulate json data so that we can choose between the default Map based representation and a fast byte buffer representation. It doesn't know or care about the structure of the data, but for performance it will care about how it's instantiated. It also cares about serialization.

Meaning will be attached to the data via extension types over maps/lists in the data; those maps and lists might be standard types or byte buffer backed special types, depending on the representation being used.

The fuzzy parts: I'm not sure yet about mutability/copying and how that needs to be set up to support convenient and fast operation in both modes.

My current idea to flesh this out is by adding a first simple schema, generating some extension types, then we can start to see how the whole thing ties together.

pkgs/dart_model/lib/src/json_changes.dart
dave26199: ;) just keep saying "it's an efficient binary representation underneath, with deduped strings" ... I do think it will be.

If you think _this_ is bad, wait until you see "Properties", a Map from "abstract", "static", etc to bools. It's going to take up a lot of space in the human readable format. I'm really torn between correctness of schema and wanting to do anything at all to make it smaller ... it would also probably get more readable! But we can make the verbose form efficient too. Let's bikeshed that one later ;)

pkgs/dart_model/lib/src/json_changes.dart
dave26199: Yeah this was painful to write :)

I don't want to optimize prematurely though, I think a TODO is the right thing for now, it looks bad but it's not clear it's a hotspot for performance, and it should be possible to optimize in place if so.

One thing that does need figuring out is what we are doing here with mutability, if we have a good story that can help a lot, let me TODO that one too :)

The translation to (binary) JSON should be fine, I am 99% sure we will end up deduping key strings so they are small and fast, we can treat strings in paths the same.

pkgs/dart_model/lib/src/json.dart
dave26199: Ah, yes, that sounds much better, thanks ... let me give that a try.

dave26199:
Thanks! Replied to design comments now so my brain can iterate on the questions raised :) will make changes and reply fully on Monday.

pkgs/dart_model/lib/src/json.dart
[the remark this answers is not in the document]
davidmorgan: Okay, I think I made some progress :)

Renamed this class to `JsonData`, it's a blob of immutable + serializable JSON data.

You access it via `asMap` which returns a read-only `JsonMap`.

We don't actually need == / hashCode, removed those.

This fits exactly with the `JsonBuffer` exploration, where there is one class holding the byte buffer and thin classes provide views into it. It also clears up the question of immutability :)

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,113 @@
...
+  Json.fromJson(Map<String, Object?> json) : _root = json;
+
+  JsonChanges computeChangesFrom(Json previous) {
+    final updates = <Update>[];
+    final removals = <Removal>[];
+    _compute(previous._root, _root, Path([]), updates, removals);
+    return JsonChanges(updates: updates, removals: removals);
+  }
+
+  static void _compute(
```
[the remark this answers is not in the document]
davidmorgan: Done.

It's here rather than on `JsonChanges` because it will need to access internals to be efficient.

pkgs/dart_model/lib/src/json.dart
davidmorgan: Yes, added a TODO.

pkgs/dart_model/lib/src/json.dart
davidmorgan: That seems reasonably likely ... see comment elsewhere re: lists of paths vs maps.

pkgs/dart_model/lib/src/json.dart
davidmorgan: So I tried structuring as maps instead of with paths, indeed it looks better; updated the code.

It does make things more complicated: mostly that it's no longer explicit where the path ends and the update begins, it's implicit in the values.

I'm not sure yet whether it is better to have two trees, one for updates and one for removals; or to merge the trees and use a different value on the leaf node to differentiate. Updates to lists would also need something additional, I guess they are easier to differentiate using a value on the leaf node but that adds significant complexity. For now I think a simple version with maps is a good place to be, added a TODO to answer the next questions :)

pkgs/dart_model/lib/src/json_changes.dart
davidmorgan: Paths are gone following the change to maps :)

pkgs/dart_model/lib/src/json_changes.dart
davidmorgan: Paths are gone following the change to maps :)

That's not the end of the performance questions, though. Right now both creating the diff and applying changes do a lot of instantiation/copying. I think it's possible to improve that, but it's a lot more complexity than fits in this PR :)

davidmorgan:
Made changes, please see what you think :)

This now fits better with what I think is needed for the byte buffer implementation: a `JsonData` class holding the data and a thin `JsonMap` class that's a view onto it.

For "applying" changes we probably want something a lot more efficient than copying everything :) I played a bit with approaches where changes are not actually merged into the data but just checked before the underlying data is checked; that should be fast for small changes. But it's quite complicated, not for this PR :)

Thanks.

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,210 @@
...
+
+import 'package:collection/collection.dart';
+
+import 'json_changes.dart';
+
+/// Immutable, serializable JSON data.
+///
+/// Views onto the data are provided as [Map] instances, see [asMap].
+final class JsonData {
+  Map<String, Object?> _root = {};
```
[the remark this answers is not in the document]
davidmorgan: Done, and opened #2 to start using lints that would have caught this :) thanks.

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  static void _deepCopyAndCheckList(List<Object?> from, List<Object?> to) {
+    for (final value in from) {
+      to.add(_deepCopyAndCheckValue(value));
+    }
+  }
+
+  static Object? _deepCopyAndCheckValue(Object? value) {
+    if (value is Map<String, Object?>) {
```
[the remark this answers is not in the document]
davidmorgan: Very nice, thanks. Done :)

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,210 @@
...
+  static void _applyUpdates(
+      {required Map<String, Object?> root,
+      required Map<String, Object?> updates}) {
+    updates.forEach((key, value) {
+      if (value is Map<String, Object?>) {
+        if (root[key] is! Map<String, Object?>) {
+          root[key] = <String, Object?>{};
+        }
+        _applyUpdates(
+            root: (root[key] ??= <String, Object?>{}) as Map<String, Object?>,
```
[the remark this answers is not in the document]
davidmorgan: I didn't think that would work, but it does :) thanks!

pkgs/dart_model/lib/src/json.dart
```diff
@@ -0,0 +1,210 @@
...
+            root: root[key]! as Map<String, Object?>, removals: value);
+      } else {
+        root.remove(key);
+      }
+    });
+  }
+}
+
+/// An immutable `Map` view onto part of [JsonData].
+final class JsonMap
```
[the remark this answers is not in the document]
davidmorgan: Done.

PR #3, opened 2024-06-18
Add minimal model schema and generation of extension types based on it.

davidmorgan:
Add a schema and a generator that generates extension types based on it.

Obviously we don't have to check in the generated code but it seems useful for now that we review it.

No particular clear ideas on the best way to lay this out, e.g. the top level "schema" folder, the generator under "tool". Happy to hear any suggestions.

No test coverage for the generator yet but we get exactly one golden file test, the output, for free :) happy to add anything you think is useful at this point.

Next steps:

1) a way to combine manually written code with the generated code, to add docs and helper methods.
2) a way to construct the extension types that is compatible with representation as a byte buffer, i.e. the underlying collections need to be swapped for special purpose ones.

### DIFF
--- a/schemas/dart_model.schema.json
+++ b/schemas/dart_model.schema.json
@@ -1,0 +1,76 @@
+{
+  "$id": "https://github.com/dart-lang/macros/blob/main/schemas/dart_model.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "oneOf": [
+    {"$ref": "#/$defs/Model"}
+  ],
+  "$defs": {
+    "Annotation": {
+      "type": "object",
+      "description": "An annotation.",
+      "properties": {
+        "type": {"$ref": "#/$defs/QualifiedName"}
+      }
+    },
+    "Interface": {
+      "type": "object",
+      "description": "An interface.",
+      "properties": {
+        "annotations": {
+          "type": "array",
+          "items": {"$ref": "#/$defs/Annotation"}
+        },
+        "members": {
+          "type": "object",
+          "description": "Map of members by name.",
+          "additionalProperties": {"$ref": "#/$defs/Member"}
+        },
+        "properties": {"$ref": "#/$defs/Properties"}
+      }
+    },
+    "Member": {
+      "type": "object",
+      "description": "Member of a scope.",
+      "properties": {
+        "properties": {"$ref": "#/$defs/Properties"}
+      }
+    },
+    "Model": {
+      "type": "object",
+      "properties": {
+        "uris": {
+          "type": "object",
+          "description": "Libraries by URI.",
+          "additionalProperties": {"$ref": "#/$defs/Library"}
+        }
+      }
+    },
+    "Properties": {
+      "type": "object",
+      "description": "Set of boolean properties.",
+      "properties": {
+        "isAbstract": {"type": "boolean"},
+        "isClass": {"type": "boolean"},
+        "isGetter": {"type": "boolean"},
+        "isField": {"type": "boolean"},
+        "isMethod": {"type": "boolean"},
+        "isStatic": {"type": "boolean"},
+        "isSynthetic": {"type": "boolean"}
+      }
+    },
+    "Library": {
+      "description": "Library.",
+      "type": "object",
+      "properties": {
+        "scopes": {
+          "type": "object",
+          "description": "Scopes by name.",
+          "additionalProperties": {"$ref": "#/$defs/Interface"}
+        }
+      }
+    },
+    "QualifiedName": {
+      "type": "string"
+    }
+  }
+}
--- a/schemas/dart_model.schema.json
+++ b/schemas/dart_model.schema.json
@@ -5,9 +5,9 @@
     {"$ref": "#/$defs/Model"}
   ],
   "$defs": {
-    "Annotation": {
+    "MetadataAnnotation": {
       "type": "object",
-      "description": "An annotation.",
+      "description": "A metadata annotation.",
       "properties": {
         "type": {"$ref": "#/$defs/QualifiedName"}
       }
@@ -16,9 +16,9 @@
       "type": "object",
       "description": "An interface.",
       "properties": {
-        "annotations": {
+        "metadataAnnotations": {
           "type": "array",
-          "items": {"$ref": "#/$defs/Annotation"}
+          "items": {"$ref": "#/$defs/MetadataAnnotation"}
         },
         "members": {
           "type": "object",
@@ -37,6 +37,7 @@
     },
     "Model": {
       "type": "object",
+      "description": "Partial model of a corpus of Dart source code.",
       "properties": {
         "uris": {
           "type": "object",
--- a/tool/dart_model_generator/bin/main.dart
+++ b/tool/dart_model_generator/bin/main.dart
@@ -1,0 +1,3 @@
+import 'package:generate_dart_model/generate_dart_model.dart';
+
+void main() => DartModelGenerator().run();
--- a/tool/dart_model_generator/bin/main.dart
+++ b/tool/dart_model_generator/bin/main.dart
@@ -1,3 +1,8 @@
-import 'package:generate_dart_model/generate_dart_model.dart';
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
 
-void main() => DartModelGenerator().run();
+import 'package:generate_dart_model/generate_dart_model.dart'
+    as dart_model_generator;
+
+void main() => dart_model_generator.run();
--- a/tool/dart_model_generator/lib/generate_dart_model.dart
+++ b/tool/dart_model_generator/lib/generate_dart_model.dart
@@ -1,0 +1,136 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:dart_style/dart_style.dart';
+import 'package:json_schema/json_schema.dart';
+
+class DartModelGenerator {
+  void run() {
+    final result = <String>[];
+    final schema = JsonSchema.create(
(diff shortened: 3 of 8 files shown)


schemas/dart_model.schema.json
```diff
@@ -0,0 +1,76 @@
...
+        "isMethod": {"type": "boolean"},
+        "isStatic": {"type": "boolean"},
+        "isSynthetic": {"type": "boolean"}
+      }
+    },
+    "Library": {
+      "description": "Library.",
+      "type": "object",
+      "properties": {
+        "scopes": {
```
jakemac53, line 65: It seems weird to call this scopes and not just declarations, which could have the regular top level declarations, which may or may not introduce a "scope". Or do we intend to model top level declarations like fields/functions as "scopes" also? It might make sense, but then you also end up wanting to probably be able to model lambda scopes, block scopes, etc....

tool/dart_model_generator/bin/main.dart
```diff
@@ -0,0 +1,3 @@
+import 'package:generate_dart_model/generate_dart_model.dart';
```
jakemac53, line 1: nit: copyright

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:dart_style/dart_style.dart';
+import 'package:json_schema/json_schema.dart';
+
+class DartModelGenerator {
```
jakemac53, line 10: nit: comments

jakemac53, line 10: Does this need to be a class? It seems like really just a namespace. Consider at least moving all the methods to static methods, but I would probably move it all to top level methods.

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
...
+    final jsonType = switch (definition.type) {
+      SchemaType.object => 'Map<String, Object?> node',
+      SchemaType.string => 'String string',
+      _ => throw UnsupportedError('Schema type ${definition.type}.'),
+    };
+    result.writeln('extension type $name.fromJson($jsonType) {');
+
+    final propertyMetadatas = definition.properties.entries
+        .map((e) => _readPropertyMetadata(e.key, e.value))
+        .toList();
```
jakemac53, line 35: consider using list literals instead:

```suggestion
    final propertyMetadatas = [
      for (var e in definition.properties.entries)
        _readPropertyMetadata(e.key, e.value),
    ];
```

Or, making `_readPropertyMetadata` accept a MapEntry so that you can pass a tearoff.

Personally, I use literals wherever possible now as opposed to functional style. But, I don't know if there are any style guidelines around this.

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
...
+    if (definition.type == SchemaType.object) {
+      result.writeln('  $name({');
+      for (final property in propertyMetadatas) {
+        result.writeln(switch (property.type) {
+          PropertyType.object =>
+            '${property.elementTypeName}? ${property.name},',
+          PropertyType.bool => 'bool? ${property.name},',
+          PropertyType.string => 'string? ${property.name},',
```
jakemac53, line 43: presumably:

```suggestion
          PropertyType.string => 'String? ${property.name},',
```

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
...
+      SchemaType.object => 'Map<String, Object?> node',
+      SchemaType.string => 'String string',
+      _ => throw UnsupportedError('Schema type ${definition.type}.'),
+    };
+    result.writeln('extension type $name.fromJson($jsonType) {');
+
+    final propertyMetadatas = definition.properties.entries
+        .map((e) => _readPropertyMetadata(e.key, e.value))
+        .toList();
+    if (definition.type == SchemaType.object) {
```
jakemac53, line 36: Nit: add a comment here describing what this is generating (the non-json constructor?)

jakemac53, line 36: Should this be a switch so we can get exhaustiveness? I think it would also look nicer, and would encourage maybe breaking out some of the code into helper methods which would make it easier to follow (`_generateConstructor` etc which are more self documenting).

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
...
+      }
+      result.writeln('}) : this.fromJson({');
+      for (final property in propertyMetadatas) {
+        result.writeln('if (${property.name} != null) '
+            "'${property.name}': ${property.name},");
+      }
+      result.writeln('});');
+    } else if (definition.type == SchemaType.string) {
+      result.writeln('$name(String string) : this.fromJson(string);');
+    }
```
jakemac53, line 58: If we don't use a switch, I would add an else that throws here.

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
...
+              'node[\'${property.name}\']!'
+              ' as ${property.elementTypeName};',
+        PropertyType.bool => 'bool get ${property.name} => '
+            'node[\'${property.name}\']! as bool;',
+        PropertyType.string => 'String get ${property.name} => '
+            'node[\'${property.name}\']! as String;',
+        PropertyType.list =>
+          'List<${property.elementTypeName}> get ${property.name} => '
+              'node[\'${property.name}\']! '
+              'as List<${property.elementTypeName}>;',
```
jakemac53, line 75: Can `property.elementTypeName` be `String`? If so, I think this cast will fail (it will be a `List<Object?>`).

Actually, I don't think this can work at all since the extensions are on a `Map<String, Object?>`? And this is a `List<Object?>` presumably?

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -0,0 +1,136 @@
...
+        PropertyType.map =>
+          'Map<String, ${property.elementTypeName}> get ${property.name} => '
+              'node[\'${property.name}\']! '
+              'as Map<String, ${property.elementTypeName}>;',
```
jakemac53, line 79: Same issue here with the value types, pretty sure this cast will fail

tool/dart_model_generator/lib/generate_dart_model.dart
davidmorgan: Added coverage to `model_test.dart`, by adding a metadata annotation, "field" accesses, and serialization. It works with the data created in the test but not after deserialization :)

Good catch, thanks! Changed to use "cast()".

tool/dart_model_generator/lib/generate_dart_model.dart
davidmorgan: Yup yup, added test coverage and changed to use "cast". Thanks.

tool/dart_model_generator/lib/generate_dart_model.dart
davidmorgan: Done :) thanks.

davidmorgan (commented):
Thanks! Ready for another look please.

davidmorgan:
Added analysis options, filed https://github.com/dart-lang/sdk/issues/56047 because it doesn't work in VSCode (for me) :)

schemas/dart_model.schema.json
```diff
@@ -0,0 +1,77 @@
...
+      "description": "Member of a scope.",
+      "properties": {
+        "properties": {"$ref": "#/$defs/Properties"}
+      }
+    },
+    "Model": {
+      "type": "object",
+      "description": "Partial model of a corpus of Dart source code.",
+      "properties": {
+        "uris": {
```
jakemac53, line 42: I don't understand what this property is for?

davidmorgan: Hmm actually maybe I did have a good reason: to avoid forcing that "things that are not libraries but go in files" end up in a different map, when probably they should go in the same map.

So maybe the Library type should be called File. URI -> file -> then semantics go on top of that.

schemas/dart_model.schema.json
```diff
@@ -0,0 +1,76 @@
...
+          "additionalProperties": {"$ref": "#/$defs/Library"}
+        }
+      }
+    },
+    "Properties": {
+      "type": "object",
+      "description": "Set of boolean properties.",
+      "properties": {
+        "isAbstract": {"type": "boolean"},
+        "isClass": {"type": "boolean"},
```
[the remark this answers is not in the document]
jakemac53: Ok, should we add at least the known things here? isEnum, isMixin, isExtension, isExtensionType, isTypeAlias, etc?

Also note that `isMixin` is a weird one - we have the `mixin` keyword as well as the `mixin` declaration which are different. Also, is a proper mixin a class?

What if we had a single enum field for the "kind" of declaration, I think I would prefer that to N booleans.

schemas/dart_model.schema.json
jakemac53: Functions definitely introduce a scope - their parameters should be resolved in the function scope (so that they can resolve to the parameters and not top level members, as an example). So, definitely more than just interfaces are at play here.

jakemac53 (approved):

schemas/dart_model.schema.json
jakemac53: "things that are not libraries but go in files" What is an example of this?

schemas/dart_model.schema.json
davidmorgan: Yes, good point. I was excluding those from "Scope", so it really isn't a good name.

I expect "Interface" to be followed by more types.

Let's pull the discussion on this and "uris" to #10

schemas/dart_model.schema.json
davidmorgan: Part files, library augmentations, new types of thing that we dream up :)

I suggest we pull this plus "scopes" discussion to #10

PR #4, opened 2024-06-18
Add LazyMap, JsonBuffer and tests.

davidmorgan:
We don't actually need to use this yet, so the implementation details are not super important.

But it _is_ important that we leave room to build the JSON data in an efficient way, I figured having this merged would help.

The `LazyMap` is the important piece: if you put data in a `LazyMap` it can be written efficiently to the buffer, because the number of keys is known so the map can be written eagerly meaning the buffer is ready to receive the actual keys+values as they are computed.

When I was working on the tree diffing code I realized that unfortunately a tree diff can't be created as a `LazyMap`. The diff algorithm has no way to compute the keys near the root before it finishes computing some leaves. So I think for quickly serializing tree diffs we'll need something smarter: a different type of `LazyMap` that can emit an `Iterable<(String, Object?)>` as it discovers keys and values. Fortunately, I don't think that complexity will leak into the rest of the data model so it should not be needed for code that is just building models.

### DIFF
--- a/pkgs/dart_model/lib/src/json_buffer.dart
+++ b/pkgs/dart_model/lib/src/json_buffer.dart
@@ -1,0 +1,390 @@
+import 'dart:collection';
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Map with lazily evaluated values.
+///
+/// This is the efficient way to accumulate data into a [JsonBuffer]. Nested
+/// maps should also be `LazyMap`.
+class LazyMap with MapMixin<String, Object?> implements Map<String, Object?> {
+  @override
+  final List<String> keys;
+  final Object? Function(String) lookup;
+
+  LazyMap(Iterable<String> keys, this.lookup)
+      : keys = UnmodifiableListView(keys.toList());
+
+  @override
+  Object? operator [](Object? key) => key is String ? lookup(key) : null;
+
+  @override
+  void operator []=(String key, Object? value) =>
+      throw UnsupportedError('LazyMap is immutable.');
+
+  @override
+  void clear() => throw UnsupportedError('LazyMap is immutable.');
+
+  @override
+  Object? remove(Object? key) =>
+      throw UnsupportedError('LazyMap is immutable.');
+}
+
+/// Bytebuffer-backed JSON data.
+///
+/// Data can be accumulated directly into the buffer which is then immediately
+/// ready to write to the wire. On the receiving side, the data can be
+/// accessed directly as a `Map<String, Object?>` without copying out of the
+/// buffer.
+///
+/// The constructor takes a [Map], but passing in a normal `Map` _does_
+/// require copying. Instead, instantiate with a [LazyMap]. The lazy map will
+/// then be evaluated into the buffer. Nested maps in the `LazyMap` must _also_
+/// use `LazyMap` otherwise it will still be necessary to copy those.
+//
+// TODO(davidmorgan): check if it's worth adding a `LazyList` for the same
+// reason.
+// TODO(davidmorgan): this is just a proof of concept to make sure `dart_model`
+// is compatible with fast JSON. Write up a design for discussion, complete
+// the implementation.
+class JsonBuffer {
+  // TODO(davidmorgan): _seenStrings and _decodedStrings are some simple
+  // optimizations; complete the design and implementation.
+  final Map<String, _Pointer> _seenStrings = {};
+  final Map<_Pointer, String> _decodedStrings = {};
+
+  /// The JSON data.
+  // TODO(davidmorgan): what's a good initial size?
+  Uint8List _buffer = Uint8List(32);
+
+  /// The next free location to write to in the buffer.
+  int _nextFree = 0;
+
+  /// Instantiates a buffer holding [map].
+  JsonBuffer(Map<String, Object?> map) {
+    _addMap(map);
+  }
+
+  /// Immutable `Map` view of the JSON data.
+  late final Map<String, Object?> asMap = JsonBufferMap._(this, 0);
+
+  /// Instantiates using previously-serialized data.
+  ///
+  /// The buffer is _not_ copied, unpredictable behavior will result if it is
+  /// mutated.
+  JsonBuffer.deserialize(this._buffer) : _nextFree = _buffer.length;
+
+  /// The JSON data.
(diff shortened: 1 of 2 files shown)


pkgs/dart_model/lib/src/json_buffer.dart
```diff
@@ -0,0 +1,390 @@
...
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Map with lazily evaluated values.
+///
+/// This is the efficient way to accumulate data into a [JsonBuffer]. Nested
+/// maps should also be `LazyMap`.
+class LazyMap with MapMixin<String, Object?> implements Map<String, Object?> {
+  @override
+  final List<String> keys;
```
jakemac53, line 11: Can we keep this as an `Iterable<String>`? Those are already unmodifiable.

pkgs/dart_model/lib/src/json_buffer.dart
```diff
@@ -0,0 +1,390 @@
...
+  final Object? Function(String) lookup;
+
+  LazyMap(Iterable<String> keys, this.lookup)
+      : keys = UnmodifiableListView(keys.toList());
+
+  @override
+  Object? operator [](Object? key) => key is String ? lookup(key) : null;
```
jakemac53, line 18: consider throwing instead of returning null, this is really an error

pkgs/dart_model/lib/src/json_buffer.dart
```diff
@@ -0,0 +1,390 @@
...
+  /// Adds a `Map` to the buffer.
+  void _addMap(Map<String, Object?> map) {
+    List<String> keys;
+    Object? Function(String) lookup;
+    if (map is LazyMap) {
+      keys = map.keys;
+      lookup = map.lookup;
+    } else {
+      keys = map.keys.toList();
+      lookup = (key) => map[key];
```
jakemac53, line 88: It would probably be good to create an extension method and use a tear-off of that here, it should be cheaper (only captures the `map` instance scope and not all the surrounding scope here).

pkgs/dart_model/lib/src/json_buffer.dart
```diff
@@ -0,0 +1,390 @@
...
+// the implementation.
+class JsonBuffer {
+  // TODO(davidmorgan): _seenStrings and _decodedStrings are some simple
+  // optimizations; complete the design and implementation.
+  final Map<String, _Pointer> _seenStrings = {};
+  final Map<_Pointer, String> _decodedStrings = {};
+
+  /// The JSON data.
+  // TODO(davidmorgan): what's a good initial size?
+  Uint8List _buffer = Uint8List(32);
```
jakemac53, line 57: Do we know how many entries there will be in each map? I think we probably could.

If so, and also everything is either a pointer or an int/bool, then we could know exactly what size to always allocate this to, I think?

I am not totally certain though, and of course optional values make that more interesting.

jakemac53: Or, just use a `BytesBuilder` instead of manually growing this ourselves? I think it basically does all the logic you are doing here.

pkgs/dart_model/lib/src/json_buffer.dart
```diff
@@ -0,0 +1,390 @@
...
+  late final Map<String, Object?> asMap = JsonBufferMap._(this, 0);
+
+  /// Instantiates using previously-serialized data.
+  ///
+  /// The buffer is _not_ copied, unpredictable behavior will result if it is
+  /// mutated.
+  JsonBuffer.deserialize(this._buffer) : _nextFree = _buffer.length;
+
+  /// The JSON data.
+  Uint8List serialize() => _buffer.sublist(0, _nextFree);
```
jakemac53, line 77: Note that this does a copy, is that what we want? What about `sublistView`?

pkgs/dart_model/lib/src/json_buffer.dart
```diff
@@ -0,0 +1,390 @@
...
+  /// Increases `_nextFree` accordingly. Expands the buffer if necessary.
+  void _reserve(int bytes) {
+    _nextFree += bytes;
+    while (_nextFree > _buffer.length) {
+      _expand();
+    }
+  }
+
+  /// Copies into a new buffer that's twice as big.
+  void _expand() {
```
jakemac53, line 131: This should probably take a `minSize` argument (can be a TODO). That way you can avoid multiple copies (and the whole loop above, just call `expand(_nextFree)`)

pkgs/dart_model/lib/src/json_buffer.dart
davidmorgan: Done, added a private `List` field to avoid a cast later on in `addMap`.

pkgs/dart_model/lib/src/json_buffer.dart
davidmorgan: The `Map` interface disagrees, but this is a specialist enough use case that we might as well try being strict, done. (I've never really liked `Object?` here although there are definitely arguments for it).

pkgs/dart_model/lib/src/json_buffer.dart
davidmorgan: The maps are stored as key and value pointers, and we know how many entries there will be if it's a `LazyMap`, but the keys and values also go into the buffer, along with all the nested maps, and we don't know how big those will be.

`BytesBuilder` indeed looks similar; but it doesn't allow accesses as you're building, while this does. Not sure if that matters, I'll add a TODO to consider `BytesBuilder`.

pkgs/dart_model/lib/src/json_buffer.dart
davidmorgan: Oh, that's what I thought I was doing ... thanks, done.

pkgs/dart_model/lib/src/json_buffer.dart
davidmorgan: Done.

pkgs/dart_model/lib/src/json_buffer.dart
davidmorgan: Done.

davidmorgan (commented):
Thanks, PTAL :)

jakemac53 (approved):

Issue #2, opened 2024-06-18
Set up lints, analysis presubmit, test presubmit


davidmorgan:

davidmorgan:
@jakemac53 do you happen to know a monorepo that's a good example to follow, please? :)

Always happy to hack something random, but I guess following best practices would be better ;)

jakemac53:
I would follow test or build, using mono_repo to generate the github configs

jakemac53:
We should use the new pub workspace feature too (see the dart-lang/build)

jakemac53:
@davidmorgan if you want I could set up mono_repo here pretty quickly, but don't want to step on your toes if you are already working towards getting CI set up

davidmorgan:
@jakemac53 yes please :)

PR #9, opened 2024-06-20
use pub workspaces

jakemac53:
As we add more packages, this should end up being a nicer workflow (just one pub solve, more reliable workflow for shared analysis options with includes, one analysis context)


jakemac53:
Related to https://github.com/dart-lang/macros/issues/2#issuecomment-2176957077

davidmorgan (approved):
Exciting stuff, thanks :)

PR #17, opened 2024-07-16
Inject macro implementation into analyzer and test it.

davidmorgan:
The "macro" is all hardcoded for now.


davidmorgan:
Dropping this in favour of PR with both analyzer+CFE together as there is overlap / they are pretty similar.

PR #18, opened 2024-07-18
Inject macro implementation into analyzer and CFE, test it.

davidmorgan:
The "macros" are just hardcoded for now.


davidmorgan:
Hi Jake,

Now that the SDK PRs are merged, this actually works!

There is a bit of a hack, the CFE wants the platform dill file at the right version, but it's pulled from the SDK ... I updated the SDK version to the latest dev version otherwise there is breakage due to the new `concurrent` package being missing. I expect this will break again, although not too often ... very happy to hear suggestions for a better way of running this :)

Thanks.

pkgs/_analyzer_macros/test/analyzer_test.dart
```diff
@@ -0,0 +1,148 @@
...
+  Iterable<DeclarationCode> get libraryAugmentations => {};
+
+  @override
+  Map<Identifier, Iterable<TypeAnnotationCode>> get mixinAugmentations => {};
+
+  @override
+  Iterable<String> get newTypeNames => [];
+
+  @override
+  void serialize(Serializer serializer) {}
```
jakemac53, line 140: Maybe throw unimplemented to ensure this isn't getting called?

pkgs/_cfe_macros/test/cfe_test.dart
```diff
@@ -0,0 +1,154 @@
...
+      packageConfigs = TestMacroPackageConfigs();
+      runner = TestMacroRunner();
+      macroImplementation = MacroImplementation(
+        packageConfigs: packageConfigs,
+        macroRunner: runner,
+      );
+    });
+
+    test('discovers macros, runs them, applies augmentations', () async {
+      final packagesPath = File('../../.dart_tool/package_config.json');
```
jakemac53, line 43: You could use `Isolate.packageConfig` to be a bit more robust here

pkgs/_cfe_macros/test/cfe_test.dart
```diff
@@ -0,0 +1,154 @@
...
+      final computeKernelResult = await computeKernel([
+        '--enable-experiment=macros',
+        '--no-summary',
+        '--no-summary-only',
+        '--target=vm',
+        '--dart-sdk-summary=${productPlatformDill.uri}',
+        '--output=${outputFile.path}',
+        '--source=${sourceFile.uri}',
+        '--packages-file=${packagesPath.uri}',
+        '--enable-experiment=macros',
```
jakemac53, line 57:
```suggestion
```

pkgs/_cfe_macros/test/cfe_test.dart
```diff
@@ -0,0 +1,154 @@
...
+          'lib/_internal/vm_platform_strong_product.dill');
+      if (!File.fromUri(productPlatformDill.uri).existsSync()) {
+        throw StateError('Failed to find platform dill: $productPlatformDill');
+      }
+      tempDir = Directory.systemTemp.createTempSync('cfe_test');
+
+      // Inject test macro implementation.
+      packageConfigs = TestMacroPackageConfigs();
+      runner = TestMacroRunner();
+      macroImplementation = MacroImplementation(
```
jakemac53, line 36: Is this a global in the CFE? Kind of hard to understand where the implementation is actually being set when it is done this way.

tool/set_sdk_ref
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash --
+#
+# Updates ref to SDK packages across pubspects. Usage:
+#
+#   tool/set_sdk_version <git ref>
+
+REF=$1
```
jakemac53, line 7: Nit: if you just pass `--help` to this tool that will become the ref, which seems less than ideal?

It probably doesn't matter all that much but maybe making it a dart script and using package:args would be worth doing.

pkgs/_cfe_macros/pubspec.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+dev_dependencies:
+  dart_flutter_team_lints: ^3.0.0
+  front_end: any
+  frontend_server: any
+  macros: ^0.1.2-main.4
+  test: ^1.25.0
+
+# Update by running: tool/set_sdk_version <git ref>
+dependency_overrides:
```
jakemac53, line 17: I believe all these overrides can/should exist only in the root (workspace) pubspec.yaml

PR #20, opened 2024-07-22
Build macros, run them, serve a (test) service, host, connect to it

davidmorgan:
Flesh out much of the end to end functionality:

 - Build macros (with just "dart compile exe")
 - Run macros (with Process)
 - Serve them (TCP/IP)
 - Connect to them
 - First message passed macro->host: which phases to run in

Next after this: add first introspection and augmentation RPCs/capability; then it will be an actual e2e macro.

### DIFF
--- a/pkgs/_macro_builder/lib/macro_builder.dart
+++ b/pkgs/_macro_builder/lib/macro_builder.dart
@@ -2,22 +2,110 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:dart_model/dart_model.dart';
 
+/// Builds macros.
+///
+/// TODO(davidmorgan): add a way to clean up generated files and built output.
 class MacroBuilder {
   /// Builds an executable from user-written macro code.
   ///
   /// Each `QualifiedName` in [macroImplementations] must point to a class that
   /// implements `Macro` from `package:macro`.
+  ///
+  /// The [packageConfig] must include the macros and all their deps.
   Future<BuiltMacroBundle> build(
-      Iterable<QualifiedName> macroImplementations) async {
-    // TODO(davidmorgan): implement.
-    // Generated entrypoint will instantiate all the `Macro` instances pointed
-    // to by `macroImplementations` then pass them to `MacroClient.run` in
-    // `package:_macro_client`.
-    return BuiltMacroBundle();
+      File packageConfig, Iterable<QualifiedName> macroImplementations) async {
+    final script = createBootstrap(macroImplementations.toList());
+
+    return await MacroBuild(packageConfig, script).build();
+  }
+
+  /// Creates the entrypoint script for [macros].
+  static String createBootstrap(List<QualifiedName> macros) {
+    final script = StringBuffer();
+    for (var i = 0; i != macros.length; ++i) {
+      final macro = macros[i];
+      // TODO(davidmorgan): pick non-clashing prefixes.
+      script.writeln("import '${macro.uri}' as m$i;");
+    }
+    script.write('''
+import 'dart:convert';
+
+import 'package:_macro_client/macro_client.dart';
+import 'package:macro_service/macro_service.dart';
+
+void main(List<String> arguments) {
+   MacroClient.run(
+      endpoint: HostEndpoint.fromJson(json.decode(arguments[0])),
+      macros: [''');
+    for (var i = 0; i != macros.length; ++i) {
+      final macro = macros[i];
+      script.write('m$i.${macro.name}()');
+      if (i != macros.length - 1) script.write(', ');
+    }
+    script.writeln(']);');
+    script.writeln('}');
+    return script.toString();
   }
 }
 
 /// A bundle of one or more macros that's ready to execute.
-class BuiltMacroBundle {}
+class BuiltMacroBundle {
+  // TODO(davidmorgan): other formats besides executable.
+  final String executablePath;
+
+  BuiltMacroBundle(this.executablePath);
+}
+
+/// A single build.
+class MacroBuild {
+  final File packageConfig;
+  final String script;
+  final Directory workspace =
+      Directory.systemTemp.createTempSync('macro_builder');
+
+  /// Creates a build for [script] with [packageConfig], which must have all
+  /// the needed deps.
+  MacroBuild(this.packageConfig, this.script);
+
+  /// Runs the build.
+  ///
+  /// Throws on failure to build.
+  Future<BuiltMacroBundle> build() async {
+    final scriptFile = File.fromUri(workspace.uri.resolve('bin/main.dart'));
+    scriptFile.parent.createSync(recursive: true);
+    scriptFile.writeAsStringSync(script.toString());
+
+    final targetPackageConfig =
+        File.fromUri(workspace.uri.resolve('.dart_tool/package_config.json'));
+    targetPackageConfig.parent.createSync(recursive: true);
+    targetPackageConfig
+        .writeAsStringSync(_makePackageConfigAbsolute(packageConfig));
+
+    // See package:analyzer/src/summary2/kernel_compilation_service.dart for an
+    // example of compiling macros using the frontend server.
+    //
+    // For now just use the command line.
+
+    final result = Process.runSync('dart', ['compile', 'exe', 'bin/main.dart'],
+        workingDirectory: workspace.path);
+    if (result.exitCode != 0) {
+      throw StateError('Compile failed: ${result.stderr}');
+    }
+
+    return BuiltMacroBundle(
+        File.fromUri(scriptFile.parent.uri.resolve('main.exe')).path);
+  }
+
+  /// Returns the contents of [pubspec] with relative paths replaced to
+  /// absolute paths, so the pubspec will work from any location.
+  String _makePackageConfigAbsolute(File pubspec) {
+    final root = pubspec.parent.parent.absolute.uri;
+    return pubspec
+        .readAsStringSync()
+        .replaceAll('"rootUri": "../', '"rootUri": "$root');
+  }
+}
--- a/pkgs/_macro_builder/lib/macro_builder.dart
+++ b/pkgs/_macro_builder/lib/macro_builder.dart
@@ -6,6 +6,8 @@ import 'dart:io';
 
 import 'package:dart_model/dart_model.dart';
 
+import 'src/bootstrap.dart';
+
 /// Builds macros.
 ///
 /// TODO(davidmorgan): add a way to clean up generated files and built output.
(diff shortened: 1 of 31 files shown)


pkgs/_macro_builder/lib/macro_builder.dart
```diff
@@ -2,22 +2,110 @@
...
+      throw StateError('Compile failed: ${result.stderr}');
+    }
+
+    return BuiltMacroBundle(
+        File.fromUri(scriptFile.parent.uri.resolve('main.exe')).path);
+  }
+
+  /// Returns the contents of [pubspec] with relative paths replaced to
+  /// absolute paths, so the pubspec will work from any location.
+  String _makePackageConfigAbsolute(File pubspec) {
```
jakemac53, line 105: This is a package config not a pubspec right?

pkgs/_macro_builder/lib/macro_builder.dart
```diff
@@ -2,22 +2,110 @@
...
+    final root = pubspec.parent.parent.absolute.uri;
```
jakemac53, line 106: This isn't really a viable option, package configs can be anywhere.

I would replace this implementation with one that resolves all URIs relative to the file itself. This implementation has a lot of real world failure modes.

pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -2,12 +2,36 @@
...
+import 'dart:convert';
+import 'dart:io';
+
 import 'package:macro/macro.dart';
 import 'package:macro_service/macro_service.dart';
 
+/// Runs macros, connecting them to a macro host.
 class MacroClient {
+  final Iterable<Macro> macros;
+  final Socket socket;
```
jakemac53, line 14: It isn't clear to me why these are instance variables, or why this class exists, as the state isn't used at all after the constructor.

The lifecycle of the `socket` is also confusing to me here, who is in charge of closing it? When and how does it get closed?

pkgs/_macro_client/test/macro_client_test.dart
```diff
@@ -0,0 +1,26 @@
...
+
+import 'package:_macro_client/macro_client.dart';
+import 'package:_test_macros/declare_x_macro.dart';
+import 'package:macro_service/macro_service.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group(MacroClient, () {
+    test('connects to service', () async {
+      final serverSocket = await ServerSocket.bind('localhost', 0);
```
jakemac53, line 16:
```suggestion
      final serverSocket = await ServerSocket.bind('localhost', 0);
      addTeardown(serverSocket.close);
```

pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -2,26 +2,91 @@
...
 import 'package:macro_service/macro_service.dart';
 
-class MacroHost {
+/// Hosts macros: builds them, runs them, serves the macro service.
+///
+/// Tools that want to support macros, such as the Analyzer and the CFE, can
+/// do so by running a `MacroHost` and providing their own `MacroService`.
+class MacroHost implements MacroService {
   final MacroServer macroServer;
+  final ListOfServices services;
```
jakemac53, line 19: This seems to be more of a higher level concept, sort of a hierarchical registry of services? I think there is probably a better name, maybe `ServiceRegistry serviceRegistry`? That would convey better that this is one service, with a pointer back to the registry of all services.

Or possibly restructure this so it doesn't add itself in the constructor, and instead in `serve` below you add it manually, I think I probably prefer that option (it doesn't seem like this list of services is needed after the constructor so keeping a back reference around after that wrong).

Another option could be just giving each `MacroService` a `MacroService? child;` field, and dropping the list of services wrapper entirely. The default implementation of `handle` would just call it on the child or throw. And each service would be required to call `super.handle(...)` if they don't handle a request.

pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -2,26 +2,91 @@
...
+  /// Determines which phases the macro implemented at [name] runs in.
+  Future<Set<int>> queryMacroPhases(
+      File packageConfig, QualifiedName name) async {
+    if (_macroPhases != null) return _macroPhases!;
+    final macroBundle = await macroBuilder.build(packageConfig, [name]);
+    macroRunner.run(macroBundle: macroBundle, endpoint: macroServer.endpoint);
+    // TODO(davidmorgan): wait explicitly for the MacroStartedRequest to
+    // arrive, remove this hard-coded wait.
+    await Future<void>.delayed(const Duration(seconds: 2));
+    return _macroPhases!;
```
jakemac53, line 61: The implicit initialization of this field seems pretty error prone. It isn't clear how/what is initializing it. I would see if instead of making it an implicit thing, you could make it an explicit return values of whichever thing initializes it, and assign it that way.

pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -2,26 +2,91 @@
...
+  }
 
-  static Future<MacroHost> serve({required MacroHostService service}) async {
-    final server = MacroServer(service: service);
-    final endpoint = await server.serve();
-    return MacroHost._(server, endpoint);
+  /// Determines which phases the macro implemented at [name] runs in.
+  Future<Set<int>> queryMacroPhases(
+      File packageConfig, QualifiedName name) async {
+    if (_macroPhases != null) return _macroPhases!;
```
jakemac53, line 55: It looks to me like there is a race condition here. Probably this should be a future (possibly a nullable future).

pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -2,26 +2,91 @@
...
-  // check if an annotation is linked to a macro, run a macro and ask it to
-  // produce augmentations.
+  /// Handle requests that are for the host.
+  @override
+  Future<Object?> handle(Object request) async {
+    // TODO(davidmorgan): don't assume the type. Return `null` for types
+    // that should be passed through to the service that was passed in.
+    final macroStartedRequest =
+        MacroStartedRequest.fromJson(request as Map<String, Object?>);
+    _macroPhases = macroStartedRequest.macroDescription.runsInPhases.toSet();
```
jakemac53, line 71: Yeah I think just completing a completer here would be better for sure (and fix the issue above)

pkgs/_macro_runner/lib/macro_runner.dart
```diff
@@ -2,9 +2,19 @@
...
 import 'package:_macro_builder/macro_builder.dart';
 import 'package:macro_service/macro_service.dart';
 
+/// Runs macros.
+///
+/// TODO(davidmorgan): support shutdown/cleanup.
 class MacroRunner {
-  void run(BuiltMacroBundle macro, HostEndpoint endpoint) {}
+  /// Runs [macroBundle] connected to [endpoint].
+  void run(
```
jakemac53, line 16: It seems a bit odd to not have this return a Future - maybe just call it `start`? It also seems like it should just be a top level method.

pkgs/_macro_runner/test/macro_runner_test.dart
```diff
@@ -0,0 +1,34 @@
...
+  group(MacroRunner, () {
+    test('runs macros', () async {
+      final builder = MacroBuilder();
+      final bundle =
+          await builder.build(File.fromUri(Isolate.packageConfigSync!), [
+        QualifiedName(
+            'package:_test_macros/declare_x_macro.dart#DeclareXImplementation')
+      ]);
+
+      final serverSocket = await ServerSocket.bind('localhost', 0);
```
jakemac53, line 24: Probably need to close this as well

pkgs/_macro_server/lib/macro_server.dart
```diff
@@ -2,15 +2,47 @@
...
+  ///
+  /// TODO(davidmorgan): other transports besides TCP/IP.
+  static Future<MacroServer> serve({required MacroService service}) async {
+    final serverSocket = await ServerSocket.bind('localhost', 0);
+    return MacroServer._(
+        service, HostEndpoint(port: serverSocket.port), serverSocket);
+  }
+
+  void _handleConnections() {
+    serverSocket.forEach(_handleConnection);
```
jakemac53, line 30: This could probably be inlined above

pkgs/_macro_server/test/macro_server_test.dart
```diff
@@ -0,0 +1,40 @@
...
+class TestMacroHostService implements MacroService {
+  final StreamController<MacroStartedRequest> _macroStartedRequestsController =
+      StreamController();
+  Stream<MacroStartedRequest> get macroStartedRequests =>
+      _macroStartedRequestsController.stream;
+
+  @override
+  Future<Object> handle(Object request) async {
+    _macroStartedRequestsController.add(request as MacroStartedRequest);
+    return MacroStartedResponse() as Object;
```
jakemac53, line 38: Why is this cast necessary?

pkgs/_test_macros/lib/declare_x_macro.dart
```diff
@@ -12,7 +12,7 @@ class DeclareX {
 // TODO(davidmorgan): this is a placeholder; make it do something, test it.
 class DeclareXImplementation implements Macro {
   @override
-  MacroDescription get description => MacroDescription(runsInPhases: [3]);
+  MacroDescription get description => MacroDescription(runsInPhases: [2]);
```
jakemac53, line 15: It seems wrong for users to be creating this description, with explicit phase numbers like this

pkgs/_macro_builder/lib/macro_builder.dart
```diff
@@ -2,22 +2,110 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
```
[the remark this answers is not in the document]
davidmorgan: For sure this will end up with multiple implementations; added a TODO.

Some of them may come from code already in the analyzer and/or CFE.

pkgs/_macro_builder/lib/macro_builder.dart
```diff
@@ -2,22 +2,110 @@
...
+  ///
+  /// The [packageConfig] must include the macros and all their deps.
   Future<BuiltMacroBundle> build(
-      Iterable<QualifiedName> macroImplementations) async {
-    // TODO(davidmorgan): implement.
-    // Generated entrypoint will instantiate all the `Macro` instances pointed
-    // to by `macroImplementations` then pass them to `MacroClient.run` in
-    // `package:_macro_client`.
-    return BuiltMacroBundle();
+      File packageConfig, Iterable<QualifiedName> macroImplementations) async {
```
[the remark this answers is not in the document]
davidmorgan: I think the key question here is about the scope of `MacroBuilder`: is it one builder for the host, one for the workspace, one for the workspace at a particular time ... wider scope allows more optimization, by which I mean reuse of (resident?) compilers and compile artifacts across builds.

I agree that `File` does not seem like the right concept, what we are really talking about here is the version solve and workspace that we need to do a build: the builder is going to have to care about both and about changes to both. There are questions of file overlays when using the analyzer and multiroot file systems.

So a lot to unpack here :) which is why I have it at the simplest+wrong thing currently. I'll add a pile of TODOs for all the above.

pkgs/_macro_builder/lib/macro_builder.dart
```diff
@@ -2,22 +2,110 @@
...
+    final script = createBootstrap(macroImplementations.toList());
+
+    return await MacroBuild(packageConfig, script).build();
+  }
+
+  /// Creates the entrypoint script for [macros].
+  static String createBootstrap(List<QualifiedName> macros) {
```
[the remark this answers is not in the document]
davidmorgan: For the "private" packages, starting with an underscore, I put implementation directly under `lib` instead of `lib/src`; my expectation is that we end up folding them entirely into some package's `src` so adding `lib/src` now just seemed like noise.

So I'm thinking of this entire file as private.

Happy to do something else though, as you like.

pkgs/_test_macros/lib/declare_x_macro.dart
davidmorgan: Yes, added a TODO on the `Macro` interface that `MacroDescription` should be populated (as far as possible) by `macro_builder` which can figure out the information and inject it in the bootstrap script, rather than asking the user macro code to do it.

pkgs/_macro_builder/lib/macro_builder.dart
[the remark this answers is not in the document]
jakemac53: Fwiw I don't think the version solve or workspace do matter - those are both encapsulated by the package config. And the package config is the only thing the compiler or analyzer should care about.

I just don't think we want File (or any dart:io type) in the public API. Even if this was just a URI I probably would be happy :).

pkgs/_macro_builder/lib/macro_builder.dart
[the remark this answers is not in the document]
jakemac53: I do think there is a difference here worth encapsulating.

If I understand correctly, MacroBuilder will be public (ie: accessible to the CFE/Analyzer). But this function is not intended to be used by either, and should ideally not be exposed.

pkgs/_macro_host/lib/macro_host.dart
jakemac53: 👍 Fine to follow up in later prs

jakemac53 (approved):

pkgs/_macro_builder/lib/macro_builder.dart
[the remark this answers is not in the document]
davidmorgan: That makes sense, moved out to `lib/src` as you originally suggested. Thanks :)

pkgs/_macro_builder/lib/macro_builder.dart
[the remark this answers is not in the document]
davidmorgan: I think we probably do have to end up worrying about version solve + workspace in order to handle finding prebuilt macros and changes to macros; it's not needed for now though.

Sure, Uri works; done. Thanks.

PR #22, opened 2024-07-24
Connect queries from macro through to host.

davidmorgan:
Add union types to JSON generation, use them for requests and responses.

### DIFF
--- a/pkgs/_macro_client/lib/macro_client.dart
+++ b/pkgs/_macro_client/lib/macro_client.dart
@@ -10,24 +10,25 @@ import 'package:dart_model/dart_model.dart';
 import 'package:macro/macro.dart';
 import 'package:macro_service/macro_service.dart';
 
-/// Runs macros, connecting them to a macro host.
+/// Local macro client which runs macros as directed by requests from a remote
+/// macro host.
 ///
 /// TODO(davidmorgan): handle shutdown and dispose.
 /// TODO(davidmorgan): split to multpile implementations depending on
 /// transport used to connect to host.
 class MacroClient {
-  late final RemoteMacroHost host;
   final Iterable<Macro> macros;
   final Socket socket;
+  late final RemoteMacroHost _host;
+  Completer<Response>? _responseCompleter;
 
   MacroClient._(this.macros, this.socket) {
+    _host = RemoteMacroHost(this);
 
-  MacroClient._(this.macros, this.socket) {
     // TODO(davidmorgan): negotiation about protocol version goes here.
 
     // Tell the host which macros are in this bundle.
     for (final macro in macros) {
-      _responseCompleter = Completer();
       _sendRequest(MacroRequest.macroStartedRequest(
           MacroStartedRequest(macroDescription: macro.description)));
     }
@@ -63,9 +64,11 @@ class MacroClient {
     switch (hostRequest.type) {
       case HostRequestType.augmentRequest:
         _sendResponse(Response.augmentResponse(
-            await macros.single.augment(host, hostRequest.asAugmentRequest)));
+            await macros.single.augment(_host, hostRequest.asAugmentRequest)));
       default:
       // Ignore unknown request.
+      // TODO(davidmorgan): make handling of unknown request types a designed
+      // part of the protocol+code, update implementation here and below.
     }
     final response = Response.fromJson(jsonData);
     switch (response.type) {
@@ -83,6 +86,15 @@ class MacroClient {
 }
 
 /// [Host] that is connected to a remote macro host.
+///
+/// Wraps `MacroClient` exposing just what should be available to the macro.
+///
+/// This gets passed into user-written macro code, so fields and methods here
+/// can be accessed by the macro code if they are public, even if they are not
+/// on `Host`, via dynamic dispatch.
+///
+/// TODO(language/issues/3951): follow up on security implications.
+///
 class RemoteMacroHost implements Host {
   final MacroClient _client;
 
--- a/pkgs/_macro_client/lib/macro_client.dart
+++ b/pkgs/_macro_client/lib/macro_client.dart
@@ -2,9 +2,11 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:dart_model/dart_model.dart';
 import 'package:macro/macro.dart';
 import 'package:macro_service/macro_service.dart';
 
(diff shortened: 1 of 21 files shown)


pkgs/macro_service/lib/src/macro_service.g.dart
```diff
@@ -40,6 +40,32 @@ extension type HostEndpoint.fromJson(Map<String, Object?> node) {
...
+  augmentRequest;
+}
+
+extension type HostRequest.fromJson(Map<String, Object?> node) {
+  static HostRequest augmentRequest(AugmentRequest augmentRequest) =>
+      HostRequest.fromJson(
+          {'type': 'AugmentRequest', 'value': augmentRequest.node});
+  HostRequestType get type {
+    switch (node['type'] as String) {
+      case 'AugmentRequest':
```
jakemac53, line 54: nit: We should consider serializing these using the enum value instead of the String value.

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -128,13 +132,70 @@ String _generateExtensionType(String name, JsonSchema definition) {
...
+/// and `asFoo` getters that "cast" to each type.
+///
+/// On the wire the union type is: `{"type": <name>, "value": <value>}`
+String _generateUnion(String name, List<JsonSchema> oneOf) {
+  final result = StringBuffer();
+  final types =
+      oneOf.map((s) => _refName(s.schemaMap![r'$ref'] as String)).toList();
+
+  // TODO(davidmorgan): add description(s).
+  result.writeln('enum ${name}Type {');
```
jakemac53, line 151: nit: consider using cascades for these

pkgs/macro_service/lib/src/macro_service.g.dart
davidmorgan: Quite possibly, what are the benefits of using the enum value?

I think I liked the name with initial upper case letter more as it matches the class name, i.e. it seems to make more sense that we refer to `AugmentRequest` on the wire than `augmentRequest`, which is what we'd get from the enum value.

Other than that I don't remember any particularly strong reason for it being like it is now :)

Thanks.

tool/dart_model_generator/lib/generate_dart_model.dart
davidmorgan: Done.

pkgs/dart_model/lib/src/dart_model.g.dart
```diff
@@ -49,6 +49,18 @@ extension type Interface.fromJson(Map<String, Object?> node) {
...
+/// Library.
+extension type Library.fromJson(Map<String, Object?> node) {
+  Library({
+    Map<String, Interface>? scopes,
+  }) : this.fromJson({
+          if (scopes != null) 'scopes': scopes,
+        });
+
+  /// Scopes by name.
+  Map<String, Interface> get scopes => (node['scopes'] as Map).cast();
```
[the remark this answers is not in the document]
jakemac53: Oh right, because the _values_ need to get cast to a `Map<String, Object?>` as well.

pkgs/macro_service/lib/src/macro_service.g.dart
jakemac53: Just because an integer is a lot smaller

jakemac53: In general maybe this is a fallout of using json schema... but we are talking about significantly larger payloads it seems like in general if we can't optimize things like this.

Maybe, in the binary form we can actually serialize these differently and use the index, and in the regular JSON form we don't?

pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -36,21 +43,54 @@ class MacroClient {
...
+          _responseCompleter!.complete(response);
+          _responseCompleter = null;
+        }
+    }
   }
 }
 
 /// [Host] that is connected to a remote macro host.
-class RemoteMacroHost implements Host {}
+class RemoteMacroHost implements Host {
```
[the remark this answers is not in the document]
jakemac53: As long as the client only sees the more restricted interface I think it would be fine? Yes, you are correct people could use `dynamic` to reach into the details of the class though. But at least right now I am not concerned about that.

pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -2,24 +2,31 @@
...
 
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:dart_model/dart_model.dart';
 import 'package:macro/macro.dart';
 import 'package:macro_service/macro_service.dart';
 
 /// Runs macros, connecting them to a macro host.
```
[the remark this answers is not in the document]
davidmorgan: Done.

pkgs/_macro_client/lib/macro_client.dart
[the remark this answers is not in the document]
davidmorgan: Yes, it's only needed from a security point of view.

For our own benefit I think it's nice to start splitting into client implementation details vs what is provided to user code, although it would be nicer still without the "this" wrapper :) probably it ends up moving out and being a clearer boundary.

pkgs/macro_service/lib/src/macro_service.g.dart
davidmorgan: Ah, got it.

As far as the schema goes I think we definitely want the string, we don't want to mess with enum ordering breaking the protocol :)

Re: the binary format, optimizing for strings is anyway a concern, pretty sure we'll end up encoding references to a repeated string as an ID, it's easy enough to do and it saves doing the UTF8 conversion lots of times. So that should cover this case, too.

davidmorgan (commented):
Thanks, PTAL.

pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -36,21 +43,54 @@ class MacroClient {
...
-    final response = await macros.single.augment(host, augmentRequest);
-    _send(response.node);
+    final jsonData = json.decode(request) as Map<String, Object?>;
+    final hostRequest = HostRequest.fromJson(jsonData);
+    switch (hostRequest.type) {
+      case HostRequestType.augmentRequest:
+        _sendResponse(Response.augmentResponse(
+            await macros.single.augment(host, hostRequest.asAugmentRequest)));
+      default:
+      // Ignore unknown request.
```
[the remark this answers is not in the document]
jakemac53: If the deserialization could give you the "unknown" type when it sees something it doesn't recognize, then that could help to ensure you are handling all the known cases at least instead of losing any exhaustiveness checking. But either way 🤷

jakemac53 (approved):

pkgs/_macro_client/lib/macro_client.dart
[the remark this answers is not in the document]
davidmorgan: Yes, deserialization gives you the "unknown" type :) ... but I think if we need to use `default` rather then explicitly checking `unknown` in switch statements, so that the switch stays exhaustive when a new type is added. Otherwise it will cause compile errors in existing code to add a new type.

Now that I think about it this is a thing macro authors could run into, we may need to give this some thought ... will file an issue.

Issue #8, opened 2024-06-19
Add strict schemas


davidmorgan:
The default schema is very loose, because JSON allows new things everywhere.

Add strict schemas that don't match all future versions, just exactly the current version.

jakemac53:
What would the idea be with these? How would we deal with SDK versioning and protocol support?

davidmorgan:
A strict schema matches as closely as possible what one specific version of the code is supposed to produce, without allowing for future changes.

So it's mostly useful for e2e tests :) to check you are actually producing the JSON you think you're producing.

Otherwise, you can put all the data under "properteis" instead of "properties" and the (loose) schema matches fine :)

There's considerable overlap with the test coverage you get from e2e tests using code generated from the schema; except that the schema can be a little stricter, it will catch a test merely _sending_ incorrect data whereas via the generated code you would have to have code that tries to _use_ it.

Probably we start with the code tests, we need those anyway, and consider adding strict schemas if it seems worth doing.

The other advantage of strict schemas, maybe, is you can conveniently compare multiple versions; whereas with codegen-based tests it's not super convenient to generate multiple versions. But, not sure what we would actually use that for :)

tl;dr: I think this is _mostly_ obsoleted by codegen from the schema.

PR #26, opened 2024-08-07
Fix unions schema to match JSON.

davidmorgan:
With a note that we should actually do https://github.com/dart-lang/macros/issues/8 so that this kind of mistake is caught automatically :)

### DIFF
--- a/schemas/macro_service.schema.json
+++ b/schemas/macro_service.schema.json
@@ -1,8 +1,15 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "oneOf": [
-    {"$ref": "#/$defs/AugmentRequest"},
-    {"$ref": "#/$defs/AugmentResponse"}
+  "anyOf": [
+    {
+      "$ref": "#/$defs/HostRequest"
+    },
+    {
+      "$ref": "#/$defs/MacroRequest"
+    },
+    {
+      "$ref": "#/$defs/Response"
+    }
   ],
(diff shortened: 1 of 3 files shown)


schemas/macro_service.schema.json
```diff
@@ -38,9 +39,9 @@
     },
     "HostRequest": {
       "$description": "A request sent from host to macro.",
-      "oneOf": [
-        {"$ref": "#/$defs/AugmentRequest"}
-      ]
+      "properties": {"type": {"type": "string"}, "value": {"oneOf": [
+          {"$ref": "#/$defs/AugmentRequest"}
+      ]}}
```
jakemac53, line 44: nit: can we format these so they are easier to understand, for objects with a single key/value or lists with a single entry I am OK with collapsing them on one line, but when it gets nested it gets pretty confusing

```suggestion
      "properties": {
          "type": {"type": "string"},
          "value": {
              "oneOf": [
                  {"$ref": "#/$defs/AugmentRequest"}
              ]
          }
      }
```

jakemac53 (approved):

schemas/macro_service.schema.json
davidmorgan: Yeeeah I was in two minds about this, it's awkward to collapse but it's also awkward expanded as it's boilerplate :)

Switched to auto formatting the whole file, which is probably what we end up doing/enforcing anyway. Thanks.

PR #27, opened 2024-08-07
Differentiate types of request explicitly, remove ListOfService.

davidmorgan:


pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -15,31 +15,23 @@ import 'package:macro_service/macro_service.dart';
...
-  static Future<MacroHost> serve({required HostService service}) async {
-    final listOfServices = ListOfServices();
-    listOfServices.services.add(service);
-    final server = await MacroServer.serve(service: listOfServices);
-    return MacroHost._(server, listOfServices);
+  /// Starts a macro host with introspection queries handled by [queryService].
+  static Future<MacroHost> serve({required QueryService queryService}) async {
+    final result = MacroHost._(queryService);
+    final server = await MacroServer.serve(service: result);
+    result._macroServer = server;
```
jakemac53, line 33: This is definitely still sending up red flags for me, assigning manually fields like this after construction is something I would always try to avoid. I get that the cyclic dependency sort of requires it here, but I would still like to see that cycle broken.

pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -15,31 +15,23 @@ import 'package:macro_service/macro_service.dart';
...
-  /// TODO(davidmorgan): make this split clearer, it should be in the protocol
-  /// definition somewhere which requests the host handles.
-  static Future<MacroHost> serve({required HostService service}) async {
-    final listOfServices = ListOfServices();
-    listOfServices.services.add(service);
-    final server = await MacroServer.serve(service: listOfServices);
-    return MacroHost._(server, listOfServices);
+  /// Starts a macro host with introspection queries handled by [queryService].
+  static Future<MacroHost> serve({required QueryService queryService}) async {
+    final result = MacroHost._(queryService);
```
jakemac53, line 31: nit: just call this `host` or something, it isn't really a "result" (there is no query or anything, its just a constructor call).

jakemac53 (approved):

pkgs/_macro_host/lib/macro_host.dart
davidmorgan: Done.

pkgs/_macro_host/lib/macro_host.dart
davidmorgan: The circle is due to the class doing too much: it's the host and the host service.

Split out the host service and now it looks much better, thanks :)

PR #28, opened 2024-08-07
Use package:_macro_host in _analyzer_macros.

davidmorgan:
This now actually builds and runs the test macro via a host injected into the analyer.

### DIFF
--- a/pkgs/_analyzer_macros/lib/macro_implementation.dart
+++ b/pkgs/_analyzer_macros/lib/macro_implementation.dart
@@ -1,0 +1,173 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:_macro_host/macro_host.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
+import 'package:dart_model/dart_model.dart';
+import 'package:macro_service/macro_service.dart';
+import 'package:macros/macros.dart';
+// ignore: implementation_imports
+import 'package:macros/src/executor.dart' as injected;
+
+/// Injected macro implementation for the analyzer.
+class MacroImplementation implements injected.MacroImplementation {
+  final Uri packageConfig;
+  final Map<String, String> macroImplByName;
+  final MacroHost _host;
+
+  MacroImplementation._(this.packageConfig, this.macroImplByName, this._host);
+
+  /// Starts the macro host.
+  ///
+  /// [packageConfig] is the package config of the workspace of the code being
+  /// edited.
+  ///
+  /// [macroImplByName] identifies macros, it's a placeholder until we identify
+  /// macros using package config. Keys are macro annotation qualified names
+  /// (`uri#name`) and values are macro implementation qualified names.
+  static Future<MacroImplementation> start({
+    required Uri packageConfig,
+    required Map<String, String> macroImplByName,
+  }) async =>
+      MacroImplementation._(
+          packageConfig,
+          macroImplByName,
+          await MacroHost.serve(
+              macroImplByName: macroImplByName,
+              queryService: AnalyzerQueryService()));
+
+  @override
+  injected.MacroPackageConfigs get packageConfigs =>
+      AnalyzerMacroPackageConfigs(this);
+
+  @override
+  injected.MacroRunner get runner => AnalyzerMacroRunner(this);
+}
+
+class AnalyzerQueryService implements QueryService {
+  @override
+  Future<QueryResponse> handle(QueryRequest request) async {
+    return QueryResponse();
+  }
+}
+
+class AnalyzerMacroPackageConfigs implements injected.MacroPackageConfigs {
+  final MacroImplementation _impl;
+
+  AnalyzerMacroPackageConfigs(this._impl);
+
+  @override
+  bool isMacro(Uri uri, String name) =>
+      _impl._host.isMacro(_impl.packageConfig, QualifiedName('$uri#$name'));
+}
+
+class AnalyzerMacroRunner implements injected.MacroRunner {
+  final MacroImplementation _impl;
+
+  AnalyzerMacroRunner(this._impl);
+
+  @override
+  injected.RunningMacro run(Uri uri, String name) => AnalyzerRunningMacro.run(
+      _impl,
+      QualifiedName('$uri#$name'),
+      // Look up from the macro name to its implementation.
+      QualifiedName(_impl.macroImplByName['$uri#$name']!));
+}
+
+class AnalyzerRunningMacro implements injected.RunningMacro {
+  final MacroImplementation _impl;
+  final QualifiedName name;
+  final QualifiedName implementation;
+  late final Future _started;
+
+  AnalyzerRunningMacro._(this._impl, this.name, this.implementation);
+
+  // TODO(davidmorgan): should this be async, removing the need for `_started`?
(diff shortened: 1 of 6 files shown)


pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
jakemac53, line 1: nit: the `analyzer_` part of the file name seems unnecessary here

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+// ignore: implementation_imports
+import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
+import 'package:dart_model/dart_model.dart';
+import 'package:macro_service/macro_service.dart';
+import 'package:macros/macros.dart';
+// ignore: implementation_imports
+import 'package:macros/src/executor.dart' as injected;
+
+/// Injected macro implementation for the analyzer.
+class MacroImplementation implements injected.MacroImplementation, HostService {
```
jakemac53, line 15: On the flip side here, I would call this AnalyzerMacroImplementation, to avoid duplicate names. Same for all those below.

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+  final Uri packageConfig;
+  final Map<String, String> macroImplByName;
```
jakemac53, line 17: I can infer the keys are names (probably class names), but I don't know what the values are. Can you add a comment describing this a bit?

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+  ///
+  /// [macroImplByName] identifies macros, it's a placeholder until we identify
+  /// macros using package config. Keys are macro annotation qualified names
+  /// (`uri#name`) and values are macro implementation qualified names.
+  static Future<MacroImplementation> start({
+    required Uri packageConfig,
+    required Map<String, String> macroImplByName,
+  }) async {
+    final result = MacroImplementation._(packageConfig, macroImplByName);
+    result._host = await MacroHost.serve(
```
jakemac53, line 35: I see this pattern is again repeated a bunch across these implementations. I will stop harping on it but something about the design here definitely seems off, that we require so much late initialization and cyclic dependencies.

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+
+  @override
+  Future<Response?> handle(MacroRequest request) async {
+    if (request.type != MacroRequestType.queryRequest) return null;
+    return Response.queryResponse(QueryResponse());
+  }
+}
+
+class MacroPackageConfigs implements injected.MacroPackageConfigs {
+  final MacroImplementation impl;
```
jakemac53, line 54: should this be private?

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+
+  MacroPackageConfigs(this.impl);
+
+  @override
+  bool isMacro(Uri uri, String name) =>
```
jakemac53, line 59: Seems a bit weird that this is just a wrapper which calls a function by the same name, but which takes slightly different args. Can we unify the signatures? Similar patterns below

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+
+class RunningMacro implements injected.RunningMacro {
+  final MacroImplementation impl;
+  final QualifiedName name;
+  final QualifiedName implementation;
+  late final Future _started;
+
+  RunningMacro._(this.impl, this.name, this.implementation);
+
+  static RunningMacro run(MacroImplementation impl, QualifiedName name,
```
jakemac53, line 84: Can this instead return a `Future<RunningMacro>` and we can drop the `_started` field and the awaits below?

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
```diff
@@ -0,0 +1,163 @@
...
+  @override
+  Future<MacroExecutionResult> executeTypesPhase(
+      MacroTarget target, TypePhaseIntrospector typePhaseIntrospector) async {
+    await _started;
+    return MacroExecutionResult(
+        target, await impl._host.augment(name, AugmentRequest(phase: 1)));
+  }
+}
+
+/// Converts [AugmentResponse] to [injected.MacroExecutionResult].
```
jakemac53, line 119: Should we have TODOs for the empty responses here?

jakemac53 (approved):

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: Yes indeed, done :) thanks.

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: Done.

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: It's described in the comment where it's passed in to `start`: value are fully qualified name of the implementation.

This will anyway go away in the next PR when the placeholder pubspec reader is used :)

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: It's probably always correct to refactor, so please do keep mentioning :) thanks.

Here again it's due to the class doing two things; moved the host service to a separate class.

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: Done.

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: That's because the injected API assumes the package config URI is known from context, but the `_macro_host` API wants it passed in.

We could update `_macro_host` API to match, also assuming it from context, but it's not clear to me yet what the scope of one `MacroHost` will be, it only makes sense if it's going to be for a single package config.

So, maybe :)

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: Yeeeah that's what I had originally, but the `RunningMacro` method that gets injected in the analyzer is sync so this has to be sync. I added a TODO to consider changing that upstream :)

pkgs/_analyzer_macros/lib/analyzer_macro_implementation.dart
davidmorgan: This class is basically one big TODO :) added at the top.

PR #29, opened 2024-08-07
Placeholder package config reader.

davidmorgan:
So we can identify macros based on pubspecs while we wait for that to be implemented properly.

Not used yet as that would overlap with other pending PRs, but it should be easy to plug in afterwards.

### DIFF
--- a/pkgs/_macro_host/lib/src/package_config.dart
+++ b/pkgs/_macro_host/lib/src/package_config.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:dart_model/dart_model.dart';
+import 'package:package_config/package_config.dart';
+
+/// Reads a package config to determine information about macros.
+class MacroPackageConfig {
+  final PackageConfig packageConfig;
+
+  MacroPackageConfig({required this.packageConfig});
+
+  factory MacroPackageConfig.readUri(Uri uri) => MacroPackageConfig(
+      packageConfig:
+          PackageConfig.parseBytes(File.fromUri(uri).readAsBytesSync(), uri));
+
+  /// Checks whether [name] is a macro annotation.
+  ///
+  /// If so, returns the qualified name of the macro implementation.
+  ///
+  /// If not, returns `null`.
+  ///
+  /// This is a placeholder implementation until `language/3728` is
+  /// implemented. It expects macros to be marked by a comment in the
+  /// annotation package `pubspec.yaml` that looks like this:
+  ///
+  /// ```
+  /// # macro <annotation name> <implementation name>
+  /// ```
+  ///
+  /// For example:
+  ///
+  /// ```
+  /// # macro lib/declare_x_macro.dart#DeclareX package:_test_macros/declare_x_macro.dart#DeclareXImplementation
+  /// ```
+  QualifiedName? lookupMacroImplementation(QualifiedName name) {
+    var packageName = name.uri;
+    if (packageName.startsWith('package:') && packageName.contains('/')) {
+      packageName = packageName.substring('package:'.length);
+      packageName = packageName.substring(0, packageName.indexOf('/'));
+    } else {
+      // TODO(davidmorgan): support macros outside lib dirs.
+      throw ArgumentError('Name must start "package:" and have a path: $name');
+    }
+    final libraryPath =
+        'lib/${name.string.substring(name.uri.indexOf('/') + 1)}';
+
+    final matchingPackage = packageConfig.packages
+        .where((p) => p.name == packageName)
+        .toList()
+        .singleOrNull;
+
+    if (matchingPackage == null) {
+      throw StateError('Package "$packageName" not found in package config.');
+    }
+
+    // TODO(language/3728): read macro annotation identifiers from package
+    // config. Until then, check the pubsec, to simulate what that feature will
+    // do.
+    final packageUri = matchingPackage.root;
+    final pubspecUri = packageUri.resolve('pubspec.yaml');
+    final lines = File.fromUri(pubspecUri).readAsLinesSync();
+
+    final implsByLibraryQualifiedName = <String, String>{};
+    for (final line in lines) {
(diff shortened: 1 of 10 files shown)


pkgs/_macro_host/lib/src/package_config.dart
```diff
@@ -0,0 +1,79 @@
...
+  /// ```
+  /// # macro declare_x_macro.dart#DeclareX package:_test_macros/declare_x_macro.dart#DeclareXImplementation
+  /// ```
+  QualifiedName? lookupMacroImplementation(QualifiedName name) {
+    var packageName = name.uri;
+    if (packageName.startsWith('package:') && packageName.contains('/')) {
+      packageName = packageName.substring('package:'.length);
+      packageName = packageName.substring(0, packageName.indexOf('/'));
+    } else {
+      throw ArgumentError('Name must start "package:" and have a path: $name');
```
jakemac53, line 41: We should have a TODO here to look up the package name by URI in the package config, to support macros defined outside of `lib` dirs.

pkgs/_macro_host/lib/src/package_config.dart
```diff
@@ -0,0 +1,79 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:dart_model/dart_model.dart';
+import 'package:package_config/package_config.dart' as package_config;
+
+/// Reads a package config to determine information about macros.
+class PackageConfig {
```
jakemac53, line 11: Lets also call this something other than PackageConfig. Really, it is something that reads a package config and extracts information about macros from it.

pkgs/_macro_host/lib/src/package_config.dart
```diff
@@ -0,0 +1,79 @@
...
+  final Uri uri;
```
jakemac53, line 12: This should probably just take in a PackageConfig instance - and have a factory constructor or static method  which can construct that from a URI. That will make it more testable and avoid re-reading/parsing it on every call to `lookupMacroImplementation`.

pkgs/_macro_host/lib/src/package_config.dart
```diff
@@ -0,0 +1,79 @@
...
+          'Package "$packageName" not found in package config at "$uri".');
+    }
+
+    // TODO(language/3728): read macro annotation identifiers from package
+    // config. Until then, check the pubsec, to simulate what that feature will
+    // do.
+    final packageUri = matchingPackage.root;
+    final pubspecUri = packageUri.replace(
+        pathSegments:
+            packageUri.pathSegments.followedBy(['pubspec.yaml']).toList());
```
jakemac53, line 64: You should be able to just do:

```suggestion
    final pubspecUri = packageUri.resolve('pubspec.yaml');
```

pkgs/_macro_host/lib/src/package_config.dart
```diff
@@ -0,0 +1,79 @@
...
+    var packageName = name.uri;
+    if (packageName.startsWith('package:') && packageName.contains('/')) {
+      packageName = packageName.substring('package:'.length);
+      packageName = packageName.substring(0, packageName.indexOf('/'));
+    } else {
+      throw ArgumentError('Name must start "package:" and have a path: $name');
+    }
+    final packageConfig = package_config.PackageConfig.parseBytes(
+        File.fromUri(uri).readAsBytesSync(), uri);
+    final libraryQualifiedName =
```
jakemac53, line 45: nit: This is really a path not just a name. It also doesn't include the `lib/` which it should, to allow macros to be defined outside of `lib/` (and below, I would expect the full path not the lib relative path in the `# macro` comments.

pkgs/_macro_host/lib/src/package_config.dart
davidmorgan: Done.

pkgs/_macro_host/lib/src/package_config.dart
davidmorgan: I guess: MacroPackageConfig :) done.

pkgs/_macro_host/lib/src/package_config.dart
davidmorgan: Done.

pkgs/_macro_host/lib/src/package_config.dart
davidmorgan: I definitely tried that! But now I try it again it works. Weird. I wonder how I screwed up :)

Thanks.

pkgs/_macro_host/lib/src/package_config.dart
davidmorgan: Done.

davidmorgan:
Thanks!

Comments addressed, and I added another commit that actually uses the new code, PTAL.

pkgs/_macro_host/lib/src/package_config.dart
jakemac53: Possibly you didn't have a trailing `/` originally for the root URI. Afaik though the package config spec requires this trailing slash.

pkgs/_macro_host/lib/src/package_config.dart
```diff
@@ -0,0 +1,80 @@
...
+      // TODO(davidmorgan): support macros outside lib dirs.
+      throw ArgumentError('Name must start "package:" and have a path: $name');
+    }
+    final libraryPath =
+        'lib/${name.string.substring(name.uri.indexOf('/') + 1)}';
+
+    final matchingPackage = packageConfig.packages
+        .where((p) => p.name == packageName)
+        .toList()
+        .singleOrNull;
```
jakemac53, line 55: I am sure it doesn't matter all that much, but creating a closure + iterable + list here does bug me :).

Luckily, `PackageConfig` implements the `[]` operator :).

```suggestion
    final matchingPackage = packageConfig[packageName];
```

davidmorgan: Done.

davidmorgan:
Rebasing on top of the CFE PR it turns out the package config class has exactly what I need to remove the hack over there, resolving `file://` URIs based on package config. So, updated to add that and also work for the CFE :) PTAL.

jakemac53 (approved):

PR #30, opened 2024-08-07
Support string and null as schema representation

simolus3:
In addition to `"type": "object"`, it looks like the schema generator partially supports using `"type": "string"` for definitions. Those definitions are broken in unions though, as the generator for unions expects all definitions to be JSON.
This PR fixes that, and also adds support for `"type": "null"` definitions. Those are useful as part of unions if we only care about the element type. To model Dart's type hierarchy for instance, we may have an `union StaticType = InterfaceType {...} | NullableType { StaticType inner } | ... | DynamicType`. There is nothing interesting about `DynamicType` apart from knowing that we're dealing with `dynamic`, so `null` makes sense as the underlying JSON representation here.

### DIFF
--- a/pkgs/dart_model/lib/src/dart_model.g.dart
+++ b/pkgs/dart_model/lib/src/dart_model.g.dart
@@ -90,7 +90,7 @@ extension type NeverType.fromJson(Null _) {
   NeverType() : this.fromJson(null);
 }
 
-/// A Dart type of the form `T?` for an inner type `T`
+/// A Dart type of the form `T?` for an inner type `T`.
 extension type NullableType.fromJson(Map<String, Object?> node) {
   NullableType({
     StaticType? inner,
--- a/pkgs/dart_model/lib/src/dart_model.g.dart
+++ b/pkgs/dart_model/lib/src/dart_model.g.dart
@@ -85,6 +85,21 @@ extension type Model.fromJson(Map<String, Object?> node) {
   Map<String, Library> get uris => (node['uris'] as Map).cast();
 }
 
+/// Representation of the bottom type [Never].
+extension type NeverType.fromJson(Null _) {
+  NeverType() : this.fromJson(null);
+}
+
+/// A Dart type of the form `T?` for an inner type `T`
+extension type NullableType.fromJson(Map<String, Object?> node) {
+  NullableType({
+    StaticType? inner,
+  }) : this.fromJson({
+          if (inner != null) 'inner': inner,
+        });
+  StaticType get inner => node['inner'] as StaticType;
+}
+
 /// Set of boolean properties.
 extension type Properties.fromJson(Map<String, Object?> node) {
   Properties({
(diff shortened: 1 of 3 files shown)


tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -37,16 +37,45 @@ String generate(String schemaJson,
...
           File('schemas/dart_model.schema.json').readAsStringSync()));
-  for (final def in schema.defs.entries) {
-    if (def.value.oneOf.isNotEmpty) {
-      result.add(_generateUnion(def.key, def.value.oneOf));
+  final allDefinitions = <String, JsonSchema>{
+    for (final def in schema.defs.entries) def.key: def.value,
+  };
+
+  for (final MapEntry(:key, :value) in allDefinitions.entries) {
+    if (value.oneOf.isNotEmpty) {
```
davidmorgan, line 45: I just merged another PR tweaking generation for `oneOf`, because the schema was not correct, please rebase; sorry about that. It moves the `oneOf` in the schema into a property called `value`, since that's what we write on the wire.

https://github.com/dart-lang/macros/pull/26

pkgs/dart_model/lib/src/dart_model.g.dart
```diff
@@ -131,3 +141,45 @@ extension type QualifiedName.fromJson(String string) {
...
   Query() : this.fromJson({});
 }
+
+enum StaticTypeType {
+  unknown,
+  nullableType,
+  voidType;
+}
+
+extension type StaticType.fromJson(Map<String, Object?> node) {
```
davidmorgan, line 151: It would be useful to see schema change + code change + generator together, but I don't see the schema change--could you add it please?

There isn't any other test coverage for the generator besides what is in the schema, so it probably makes sense to add something to the schema even if it's just showing the new functionality. In that case please note in the description that the type is currently just there to show the union functionality :)

davidmorgan (commented):
Generally looks good, thanks, please also add schema as suggested so I can review fully.

pkgs/dart_model/lib/src/dart_model.g.dart
simolus3: I didn't mean to commit the type changes in this PR, but I've added a few types with their schema definitions to test the generator.

schemas/dart_model.schema.json
```diff
@@ -86,6 +86,19 @@
         }
       }
     },
+    "NeverType": {
+      "type": "null",
+      "description": "Representation of the bottom type [Never]."
+    },
+    "NullableType": {
+      "type": "object",
+      "description": "A Dart type of the form `T?` for an inner type `T`",
```
davidmorgan, line 95: Nitpick: please end with `.`

davidmorgan (approved):
Looks good, thanks!

davidmorgan:
Please fix lint violation

```
   info - lib/generate_dart_model.dart:60:33 - Angle brackets will be interpreted as HTML. Try using backticks around the content with angle brackets, or try replacing `<` with `&lt;` and `>` with `&gt;`. - unintended_html_in_doc_comment
```

PR #32, opened 2024-08-08
Use package:_macro_host in _cfe_macros.

davidmorgan:
This needs an upstream fix as the code I added in the CFE is passing the wrong macro name, will send it today.

The rest is mostly the same as the analyzer version, except: 1) the CFE is passing a mix of absolute URIs and package URIs, currently I just work around that in the test; and 2) the CFE breaks if empty augmentations are returned, I worked around that too with a TODO.


davidmorgan:
https://dart-review.googlesource.com/c/sdk/+/379640

davidmorgan:
BTW there is clearly a temptation here to dedupe code between the two :)

I'm not sure yet how much duplication there will be, most of it will be the query handling, a lot of which will be different for analyzer and CFE. Even with that, I guess there's a good chance we end up with a common package sharing code between the two. Probably a few PRs down the line is the time to add such a thing :)

davidmorgan:
Upstream landed, so updated to use that ref in the pubspec making this now landable :)

pkgs/_cfe_macros/lib/macro_implementation.dart
```diff
@@ -0,0 +1,174 @@
...
+import 'package:dart_model/dart_model.dart';
+// ignore: implementation_imports
+import 'package:front_end/src/macros/macro_injected_impl.dart' as injected;
+import 'package:macro_service/macro_service.dart';
+import 'package:macros/macros.dart';
+// ignore: implementation_imports
+import 'package:macros/src/executor.dart' as injected;
+
+/// Injected macro implementation for the analyzer.
+class MacroImplementation implements injected.MacroImplementation {
```
jakemac53, line 15: nit: I would rename this one also (`CfeMacroImplementation`), and rename the analyzer one also.

jakemac53 (approved):

pkgs/_cfe_macros/lib/macro_implementation.dart
davidmorgan: Done.

PR #33, opened 2024-08-09
First query via the analyzer.

davidmorgan:
Query for the fields of a class.

Check the output using a macro that just outputs the query response as a comment.

### DIFF
--- a/pkgs/_analyzer_macros/lib/query_service.dart
+++ b/pkgs/_analyzer_macros/lib/query_service.dart
@@ -1,0 +1,54 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:_macro_host/macro_host.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/summary2/linked_element_factory.dart' as analyzer;
+// ignore: implementation_imports
+import 'package:dart_model/dart_model.dart';
+import 'package:macro_service/macro_service.dart';
+// ignore: implementation_imports
+
+// Hack to access analyzer internals via the introspector that's available.
+// TODO(davidmorgan): handle lifecycle, are these equivalent, do they need
+// tracking per macro application?
+// TODO(davidmorgan): remove hack by injecting an explicit API into the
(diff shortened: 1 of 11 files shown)


pkgs/_analyzer_macros/lib/query_service.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'package:dart_model/dart_model.dart';
+import 'package:macro_service/macro_service.dart';
+// ignore: implementation_imports
+
+// Hack to access analyzer internals via the introspector that's available.
+// TODO(davidmorgan): handle lifecycle, are these equivalent, do they need
+// tracking per macro application?
+// TODO(davidmorgan): remove hack by injecting an explicit API into the
+// analyzer, once we know what's needed.
+Object? introspector;
```
jakemac53, line 18: Analyzer implementations should probably be passed in to `AnalyzerMacroImplementation.start` and then passed here.

pkgs/_test_macros/lib/query_class.dart
```diff
@@ -18,7 +20,14 @@ class QueryClassImplementation implements Macro {
 
   @override
   Future<AugmentResponse> augment(Host host, AugmentRequest request) async {
```
jakemac53, line 22: I think in the future we will want a `Future<void>` for the return type here, and there should be methods on `host` to perform various augmentations.

This will make macros more composable without needing to merge augmentation responses together.

The host probably needs to be phase specific as well. We might also want phase specific response types. And then phase specific methods that you implement with those more specific types on the parameters. Maybe `augmentTypes`, `augmentDeclarations`, etc.

jakemac53 (approved):

pkgs/_analyzer_macros/lib/query_service.dart
davidmorgan: I think we need to avoid using the analyzer instance that we currently create for testing, because in the end the way this code will be integrated it won't control creation of the analyzer instance.

So we need to get the analyzer via the injection. Just, via a better interface than the one that exists now :)

pkgs/_test_macros/lib/query_class.dart
davidmorgan: That all makes sense, thanks.

PR #35, opened 2024-08-09
Use a BytesBuilder in JsonBuffer

jakemac53:
This is an alternative to managing our own Uint8List, and will only ever copy bytes one time (when materializing the final list) instead of N times as it grows.

This would be cleaner if we separated the interface into a `JsonBufferReader` and `JsonBufferWriter` instead of trying to handle both. The reader should just take a Uint8List, which would avoid the weird calls to `serialize` in order to get at the bytes. Note however that this is still a cheap operation, it just is a very weird way to do things.

~Do we have benchmarks we can use here to evaluate any differences? My primary goal is not performance (it is maintainability), but I want to make sure it does not regress things. And for large payloads it should do less copying so there might be a side benefit.~ Added a new benchmarks in https://github.com/dart-lang/macros/pull/36

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 **/.dart_tool
 **/pubspec.lock
+**/*.exe
--- a/pkgs/_analyzer_macros/lib/macro_implementation.dart
+++ b/pkgs/_analyzer_macros/lib/macro_implementation.dart
@@ -4,6 +4,8 @@
 
 import 'package:_macro_host/macro_host.dart';
 // ignore: implementation_imports
+import 'package:analyzer/src/summary2/macro_declarations.dart' as analyzer;
+// ignore: implementation_imports
 import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
 import 'package:dart_model/dart_model.dart';
 import 'package:macro_service/macro_service.dart';
@@ -11,32 +13,26 @@ import 'package:macros/macros.dart';
 // ignore: implementation_imports
 import 'package:macros/src/executor.dart' as injected;
 
+import 'query_service.dart';
+
 /// Injected macro implementation for the analyzer.
 class AnalyzerMacroImplementation implements injected.MacroImplementation {
   final Uri packageConfig;
-  final Map<String, String> macroImplByName;
   final MacroHost _host;
 
-  AnalyzerMacroImplementation._(
-      this.packageConfig, this.macroImplByName, this._host);
+  AnalyzerMacroImplementation._(this.packageConfig, this._host);
 
   /// Starts the macro host.
   ///
   /// [packageConfig] is the package config of the workspace of the code being
   /// edited.
-  ///
-  /// [macroImplByName] identifies macros, it's a placeholder until we identify
-  /// macros using package config. Keys are macro annotation qualified names
-  /// (`uri#name`) and values are macro implementation qualified names.
   static Future<AnalyzerMacroImplementation> start({
     required Uri packageConfig,
-    required Map<String, String> macroImplByName,
   }) async =>
       AnalyzerMacroImplementation._(
           packageConfig,
-          macroImplByName,
           await MacroHost.serve(
-              macroImplByName: macroImplByName,
+              packageConfig: packageConfig,
               queryService: AnalyzerQueryService()));
 
   @override
@@ -47,13 +43,6 @@ class AnalyzerMacroImplementation implements injected.MacroImplementation {
   injected.MacroRunner get runner => AnalyzerMacroRunner(this);
 }
 
-class AnalyzerQueryService implements QueryService {
-  @override
-  Future<QueryResponse> handle(QueryRequest request) async {
-    return QueryResponse();
-  }
-}
-
 class AnalyzerMacroPackageConfigs implements injected.MacroPackageConfigs {
   final AnalyzerMacroImplementation _impl;
 
@@ -61,7 +50,7 @@ class AnalyzerMacroPackageConfigs implements injected.MacroPackageConfigs {
 
   @override
   bool isMacro(Uri uri, String name) =>
-      _impl._host.isMacro(_impl.packageConfig, QualifiedName('$uri#$name'));
+      _impl._host.isMacro(QualifiedName('$uri#$name'));
 }
 
 class AnalyzerMacroRunner implements injected.MacroRunner {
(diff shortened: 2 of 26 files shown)


jakemac53:
Based on the basic benchmark I added, this is currently about 2x slower.

My hunch is this is related to many allocations of small Uint8Lists. I will try to validate that and see if I can resolve it.

jakemac53:
I could get an some easy wins, avoiding calling `addByte` which creates single element Uint8List instances, but it is still about 50% slower.

Another alternative that occurred to me is we could compute the entire size up front. But it would involve invoking the value functions twice per value for lazy maps. I am not exactly sure what the use case will be for the lazy maps though and whether that would matter.

davidmorgan:
Fun stuff, thanks :)

With the `TODO` for BytesBuilder `addByte` performance it looks like we would do better adding out own `BytesBuilder` type if we want one to improve the code here; then it will do exactly what we want it to do :)

If possible I think we want to hit zero copies for most of the bytes. The "grow if needed" strategy saves copying everything after the last "grow"; e.g. if you set the grow factor to 5 then you get zero copies on up to 4/5ths of the bytes :) whereas with a `BytesBuilder` we still have to copy them all.

If we end up handling only one request at a time ... which I guess is the case on the client, and will be the case on the host if we can't figure out how to parallelize macros ... then maybe we end up using one large (1MB? 10MB?) buffer that we just reuse :) and so hitting zero copies unless some unusual huge request/response comes along.

I expect we end up building most of the data via lazy maps, as it is the only way to avoid building everything up in memory before serialization starts. Whether running through twice to be more efficient with memory could be faster, I don't know--if we can guess a large enough buffer size in advance then I guess it must be slower :)

Re: splitting into reader/writer, indeed it's neater if there is some split. I started with the idea of preserving the functionality of the "JSON Map" we're emulating, so you can immediately access data on it that you added. That's certainly simpler for users of the data model, so it might be worth having, but I'm not sure if we actually need it. Probably we could get a better idea there by integrating end to end, i.e. supporting binary in the e2e tests, to see some realistic usage. That would also be interesting for benchmarking.

jakemac53:
> With the `TODO` for BytesBuilder `addByte` performance it looks like we would do better adding out own `BytesBuilder` type if we want one to improve the code here; then it will do exactly what we want it to do :)

We could just use the copying bytes builder as well, which does essentially what the current implementation does. I will try it out.

> Re: splitting into reader/writer, indeed it's neater if there is some split. I started with the idea of preserving the functionality of the "JSON Map" we're emulating, so you can immediately access data on it that you added. That's certainly simpler for users of the data model, so it might be worth having, but I'm not sure if we actually need it. Probably we could get a better idea there by integrating end to end, i.e. supporting binary in the e2e tests, to see some realistic usage. That would also be interesting for benchmarking.

We could still make it so the writer is also a reader - but the reader does not support writing? I don't believe there is a use case for modifying a response you got back from the server (or the server modifying something it got back from the client).

The main use case here is the reader is best backed by just a simple Uint8List, and modification is almost certainly an error anyways.

jakemac53:
> If we end up handling only one request at a time ... which I guess is the case on the client, and will be the case on the host if we can't figure out how to parallelize macros ... then maybe we end up using one large (1MB? 10MB?) buffer that we just reuse :) and so hitting zero copies unless some unusual huge request/response comes along.

We definitely need to support multiple requests at the same time.

Also we can't re-use the `Uint8List`s anyways because `Socket.add` says you cannot modify the list. There is no lifecycle exposed so you have no idea when it is done and could be re-used.

Maybe we could re-use them after we get a response back, but not all requests have responses etc... it gets very complicated I think without adding extra handshakes. 

> If possible I think we want to hit zero copies for most of the bytes. The "grow if needed" strategy saves copying everything after the last "grow"; e.g. if you set the grow factor to 5 then you get zero copies on up to 4/5ths of the bytes :) whereas with a `BytesBuilder` we still have to copy them all.

I am not sure we would want to grow it that rapidly, but yes. If we only grow by doubling (which is pretty standard I think), then because of the duplicate copying of previous bytes you actually end up with worst case  O(n*2 - initial size) bytes copied and best case O(n - initial size) bytes copied.

Given a buffer size of 10 as an example:

- up to 10 bytes: no copying, yay!
- 11-20 bytes: copy 10 bytes
- 21-40 bytes: 10 + 20 = 30 bytes copied
- 41-80 bytes: 30 + 40 = 70 bytes copied
- 81 - 160 bytes: 70 + 80 = 150 bytes copied 
- etc

This does get better if you grow more aggressively for sure though.

> I'm not sure I follow how it's the same--won't that guarantee one copy of everything? Whereas the current implementation only copies if it has to resize.

Right, sorry you are correct it would involve extra copying when adding into the buffer. For strings we already do that but not for other stuff.

davidmorgan:
> > If we end up handling only one request at a time ... which I guess is the case on the client, and will be the case on the host if we can't figure out how to parallelize macros ... then maybe we end up using one large (1MB? 10MB?) buffer that we just reuse :) and so hitting zero copies unless some unusual huge request/response comes along.
> 
> We definitely need to support multiple requests at the same time.

I guess so; but it seems like they should be rare, because we can do wide queries and/or batch requests; so making one request at a time fast should cover pretty much everything.

> Also we can't re-use the `Uint8List`s anyways because `Socket.add` says you cannot modify the list. There is no lifecycle exposed so you have no idea when it is done and could be re-used.
> 
> Maybe we could re-use them after we get a response back, but not all requests have responses etc... it gets very complicated I think without adding extra handshakes.

I'm sure we could figure out a way if it turned out to be a performance win :) ... we should measure first to check if it's worth the effort though.

> > If possible I think we want to hit zero copies for most of the bytes. The "grow if needed" strategy saves copying everything after the last "grow"; e.g. if you set the grow factor to 5 then you get zero copies on up to 4/5ths of the bytes :) whereas with a `BytesBuilder` we still have to copy them all.
> 
> I am not sure we would want to grow it that rapidly, but yes. If we only grow by doubling (which is pretty standard I think), then because of the duplicate copying of previous bytes you actually end up with worst case O(n*2 - initial size) bytes copied and best case O(n - initial size) bytes copied.
> ...
> This does get better if you grow more aggressively for sure though.

Yes; also we could keep state about the sizes, e.g. set the starting size to the max size that was needed so far, shrink that size after a while if it's repeatedly underused.

This is all speculative really until we have end to end benchmarks showing it's worth the complexity. I do think it's worth getting binary end to end soon so we can make sure we're on a good path.

Anyway this is useful stuff to think about, thanks :)

jakemac53:
I am going to go ahead and close this for now given the performance loss - but might circle back and just try to separate out the buffer into its own class so the logic of that is separated out and individually testable.

davidmorgan:
Sounds good, thanks.

PR #36, opened 2024-08-09
add some basic JsonBuffer benchmarks

jakemac53:
In an effort to evaluate https://github.com/dart-lang/macros/pull/35 i added some very basic benchmarks here.

Note that the BytesBuilder is currently about 2x slower than the current implementation, but I will follow up in that PR regarding that.

### DIFF
--- a/pkgs/dart_model/benchmark/json_buffer.dart
+++ b/pkgs/dart_model/benchmark/json_buffer.dart
@@ -1,0 +1,41 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:typed_data';
+import 'package:benchmark_harness/benchmark_harness.dart';
+import 'package:dart_model/src/json_buffer.dart';
+
+Uint8List? serialized;
+JsonBuffer? deserialized;
+const mapSize = 1000000;
+final mapKeys = List.generate(mapSize, (i) => i.toString());
+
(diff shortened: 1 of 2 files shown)


pkgs/dart_model/benchmark/json_buffer.dart
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+class JsonBufferDeserializeBenchmark extends BenchmarkBase {
+  JsonBufferDeserializeBenchmark() : super('JsonBufferDeserialize');
+
+  @override
+  void run() {
+    // TODO: This is actually a no-op, so the benchmark is kind of pointless.
+    // Should we read all the keys/values?
```
davidmorgan, line 38: Yeah it's a fun one; when I was playing with this I tried two things: deep copy out to a normal map, and "process", which was just computing a hash code over everything.

https://github.com/dart-lang/language/blob/5527a8f2825f1b9fa69d7efce9ba6102bfa7aa14/working/macros/dart_model/testing/json_benchmark/lib/json_benchmark.dart#L69

I think "process" probably made the most sense of the two, although it's still only a rough approximation to real use.

davidmorgan (approved):
LG, thanks :)

pkgs/dart_model/benchmark/json_buffer.dart
jakemac53: Yeah I also was realizing that for the `serialize` benchmark its really the creation, not the serialization which takes up the vast majority of the time. Which is by design of course. That benchmark makes more sense though still.

PR #37, opened 2024-08-09
WIP: Add static type system

simolus3:
This adds an implementation of Darts type system, beginning with subtype queries, to the local model for macros.

The basic idea is that:

1. We want to subtype checks to be efficient ( => no roundtrips to the macro executor). For interface types, this requires knowing all transitive supertypes (which can be a recursive structure and is generally pretty large, so we don't want to transfer that in every message).
   - So, we split types into `ResolvedType`s (fully resolved) and type descriptions (lightweight, serializable).
   - todo: There needs to be an (necessarily asynchronous) API to turn type descriptions into resolved types.
2. Given that we know everything about `ResolvedType`s, we can do common type tasks on them (subtype queries, type algebra/substitutions, flattening, normalization, ...).
3. A type system implementation needs to have a reference to some predefined types taking a special role. For instance, some types are in a subtype relationship if the special type `Null` is in a subtype relationship with one of the types. This requires the `ResolvedType` for `Null` to be available to the type system implementation.
   - This approach has it as a parameter on `StaticTypeSystem`, with the intention of these types being resolved by querying the executor.

This obviously still needs a lot of work to integrate it into the CFE/analyzer (and also tests!), but I wanted to open a draft PR already to get some feedback on the general direction and API.


davidmorgan:
Thanks Simon! A great start.

One procedural note, those of us on the Dart team working on macros are very geographically distributed, so we usually only chat synchronously a few times per week. That means coherent feedback can take a while to form, I will aim to discuss with the group tomorrow, Tuesday, to make sure we are giving you the best guidance here :)

With that in mind, some initial thoughts from me.

I think most/all of this should go in `dart_model`: anything that can be useful with just the model and no other deps probably belongs there.

An important note, one hard problem that we want to solve in `dart_model` is compatibility across different language versions, including of course versions that we don't know about yet. In particular, we want that macros usually continue to work with no code changes against new SDK versions. We expect to support both incremental and breaking changes to the data model, so for example we allow "unknown" and will be able to add new types of things when it isn't breaking; but the wire format will be versioned so that if we need to switch to an entirely incompatible model we can do that. Having the type computation in the same package seems like a reasonable way to go for all this, but we'll have to think about it :)

Re: sync and async, with the query API the expectations about how a macro interacts with the host changes a bit: rather than a back and forth conversation in which the macro learns more about the code, we would prefer a small number of queries, maybe even a single query and response; this should be more efficient and means most macro code can then be synchronous. So rather than "here is a type, please resolve it" we are looking at "here is a model, please resolve it" or more likely in the initial query you already specify which types you want to come back already resolved.

In this setup I think the code of `StaticTypeSystem` ends up being a feature of `Model`; i.e. a `Model` is something that can be fully resolved, and if it is, you can ask it any questions about types synchronously. If not you can make an API call to resolve it.

So I think all the methods that manipulate resolved types end up using a `Model` for context. Possibly it gets passed around everywhere. Since the idea is that macros mostly get one `Model` that they process, maybe it could be `Zone`-global for convenience.

With this setup, I wonder if it's possible to unify the resolved and unresolved types? It's okay if an unresolved type fails with a runtime error "you should be running on a resolved model", because macros will generally know in advance that they need a resolved model and can just always ask for one.

Re: the overhead of repeating core type information, that's okay for now, we can optimize in lots of ways, in particular note that `Model` is intended to be diffable/composable so we can always subtract the repeated part from what we send and add it back later.

There is another type of "unresolved" to worry about which is "the host can't resolve this yet". For example inspection is very limited in phase 1. We might want to be able to encode a type which we know the name of but it's not resolved yet, e.g. we know it's called "int" in _this_ source file, but imports were not resolved yet so we don't know yet if it's `dart:core#int`, it's more like `?package:my_package/foo.dart#int`. Probably there's nothing concrete to do on that right now, but something to bear in mind to add as we build this out.

Disclaimer: from those working on macros, including yourself, I am probably _least_ familiar with the details of representing Dart types, since others have done it at least once (Konstantin on the analyzer, Johnni on the CFE, Jake in  package:macros). So if I say something that doesn't seem to make sense, it probably doesn't ;) and of course if anything is unclear, happy to elaborate.

Thanks :)

davidmorgan:
So, we discussed :)

Generally we like the direction, we are going to need serializable classes that can describe types for use in other places, having code on top of those to do computations about types makes sense, answering questions client-side instead of doing an RPC will be faster and fits better with the general idea of `dart_model`.

The analyzer and CFE currently implement this logic independently, and adding a third implementation obviously increases complexity. A good outcome would be if we end up with a unified set of test cases covering all three. It's okay for macros to have an "experimental/untested" implementation for a while as we work towards that.

We do need to dig a big into the question of whether we should have parallel resolved/unresolved types of whether they can/should be unified, as I mentioned in my previous comment.

Thanks :)

simolus3:
Thanks for the detailed responses!

> I think most/all of this should go in `dart_model`: anything that can be useful with just the model and no other deps probably belongs there.

A problem is that resolving subtypes (if we end up doing it client-side, which IMO makes sense) requires an initial RPC call to query special types in the hierarchy. For instance, we need access to the resolved `FutureOr` class at the moment because some subtype rules are `if ..., then S is a subtype of T iff S is a subtype of Future<T>`. Implementing this requires the ability to instantiate these special types synchronously, so we need all the information in place first.
We _could_ hardcode knowledge about these types in `package:dart_model`, but that feels a little fragile to me. If that helps keeping the implementation in one place, it could be worth it though.
Integrating them into `Model` would probably solve this.

> Having the type computation in the same package seems like a reasonable way to go for all this, but we'll have to think about it :)

I don't think the type implementation stands in the way of backwards/forwards compatibility w.r.t. the protocol. We can't answer subtype queries involving types we don't understand obviously, but there's nothing fundamentally incompatible with unknown types.

> So rather than "here is a type, please resolve it" we are looking at "here is a model, please resolve it" or more likely in the initial query you already specify which types you want to come back already resolved.

I see, that makes sense :+1: Are there design docs or anything to read on the kind of queries currently envisioned? I see that there's a query resolving a single class (which I guess should then also resolve its `thisType` in the type hierarchy), will that then be expanded to include libraries/multiple independent classes?

> With this setup, I wonder if it's possible to unify the resolved and unresolved types? It's okay if an unresolved type fails with a runtime error "you should be running on a resolved model", because macros will generally know in advance that they need a resolved model and can just always ask for one.

Interface types are the only thing where it makes a considerable difference (all other types are either singletons or structural by definition, so there's no difference), so I think we can unify these representation.
It might be interesting to have this distinction in the type level (e.g. `class ResolvedType implements UnresolvedType`), but runtime errors are probably fine too (or we could have an `bool get isUnresolved;` property on types).

> There is another type of "unresolved" to worry about which is "the host can't resolve this yet". For example inspection is very limited in phase 1.

IMO type analysis is not that useful in phase 1 either way, so there's a question of whether we want to deal with ambiguous/unresolved information either way or just outright reject type queries.

simolus3:
> With this setup, I wonder if it's possible to unify the resolved and unresolved types?

I'm now using a representation that can be converted from and to the serialized protocol without any loss of information, but is still distinct from it. It has two main benefits:

1. Using a resolved representation for type parameters. Given their recursive nature (`<T extends List<T>>`), the serialized form has to introduce a level of indirection `TypeParameter(id: 0, bound: NamedType(name: List, instantiation: [TypeParameterType(id: 0)]))`. Dealing with this mapping of available identifiers to type parameters everywhere is super annoying because we have to pass that context around whenever we deal with any types. Dart objects can have cyclic references, and using them greatly simplifies things.
2. We're able to express known types as `sealed` subtypes, which allows pattern matching over them - this is not as important as the first point but it makes the subtype check implementation much more readable, almost matching the mathematical definition in the language spec.

Importantly though, this representation has a direct mapping from and to the serialized type representation. To express type hierarchies, I've introduced a separate `TypeHierarchyEntry` definition on `Model`. When querying a model, it is supposed to return entries for all classes resolved in a model (as well as their transitive superclasses). That way, the type system implementation can use the type hierarchy to implement subtype checks without having to encode this information in static types directly (which would not make them 1:1 mappable to their serialized form).

davidmorgan:
Thanks Simon :) sorry I missed replying a few days ago.

Re: queries, we plan to develop that code soon, currently we are spending a bit of time on the binary JSON format to make sure that when we build out the query code it actually works with the wire format we want :)

Re: `sealed` subtypes and pattern matching, I think we actually need to avoid that--because adding a new subtype would break existing code. More generally, class hierarchies are fragile and so hard to evolve, basing everything on the generated JSON extension types should force that we don't get into that territory.

Which brings us squarely back to the circular references problem :) ... from the point of view of a macro, there is only _one_ `Model`, and it gets some part of it based on what it decides to query. So I think there should be a workable path where the model used to resolve types is a global variable; actually probably Zone scoped for things like running multiple macros in parallel.

How does that sound?

simolus3:
I think it should be possible for macros to compose types they have gotten from the model into new types, and then use these types for queries in the model's type system. For instance, in `package:drift_dev` (currently based on analyzer + build\_runner), I'm creating `RecordType` instances from an SQL schema and I'm then verifying that another `RecordType` instance I've received from analyzing source code is a subtype of that.
I think it should be possible to do the same thing with macros: The second `RecordType` would be queried and part of the model, but my macro has to create the first one and then compare them with the type system.

This means that there will be types that reference the model without being a part of it. Since records and function types are purely structural and don't have an identity on them, using the existing schema classes for them is not a problem. However, (at least to me) it seems to break apart with type parameters:

> Which brings us squarely back to the circular references problem :) ... from the point of view of a macro, there is only _one_ `Model`, and it gets some part of it based on what it decides to query

The problem is that we have to synthesize new intermediate types to answer subtype queries. In particular, looking at the [positional function types and named function types rules](https://github.com/dart-lang/language/blob/main/resources/type-system/subtyping.md#rules) from the language spec, the way we check the subtype relationship between function types is to create new function types pretending to have the same type parameters. That then allows comparing the individual parts (return types, arguments) making up the function type. However, this requires generating new type parameters! These are necessarily independent from the model, which is tricky to do with ids being passed around as the client would then have to mutate its model to allocate identifiers for these intermediate type parameters.

> Re: `sealed` subtypes and pattern matching, I think we actually need to avoid that--because adding a new subtype would break existing code

Well there's a private `_UnknownType` subclass, similarly to the unknown type in the union enum in generated code. So as long as we're not removing types (which would be breaking with the generated code as well), I don't see how this representation is harder to evolve.

Maybe I'm lacking vision wrt. to the one `Model` approach - if you have ideas on how we'd solve the problem of introducing new types derived from the model without being part of it, I can explore these approaches too.

davidmorgan:
> I think it should be possible for macros to compose types they have gotten from the model into new types, and then use these types for queries in the model's type system. For instance, in `package:drift_dev` (currently based on analyzer + build_runner), I'm creating `RecordType` instances from an SQL schema and I'm then verifying that another `RecordType` instance I've received from analyzing source code is a subtype of that. I think it should be possible to do the same thing with macros: The second `RecordType` would be queried and part of the model, but my macro has to create the first one and then compare them with the type system.
> 
> This means that there will be types that reference the model without being a part of it. Since records and function types are purely structural and don't have an identity on them, using the existing schema classes for them is not a problem. However, (at least to me) it seems to break apart with type parameters:
> 
> > Which brings us squarely back to the circular references problem :) ... from the point of view of a macro, there is only _one_ `Model`, and it gets some part of it based on what it decides to query
> 
> The problem is that we have to synthesize new intermediate types to answer subtype queries. In particular, looking at the [positional function types and named function types rules](https://github.com/dart-lang/language/blob/main/resources/type-system/subtyping.md#rules) from the language spec, the way we check the subtype relationship between function types is to create new function types pretending to have the same type parameters. That then allows comparing the individual parts (return types, arguments) making up the function type. However, this requires generating new type parameters! These are necessarily independent from the model, which is tricky to do with ids being passed around as the client would then have to mutate its model to allocate identifiers for these intermediate type parameters.

That does complicate things a bit :)

One way to do it would be to create a new temporary `Model` which is the base model + additions on top. Since it's a `Map` underneath this should be easy to do, the temporary `Map` returns values if it has them or delegates to the base model. Then you can throw away the temporary model when done. A stack of scopes/models and a `runScoped(Model, void Function())` method might do this nicely.

I am not going to claim that that's a _good_ design, but I think it would work? :)

> > Re: `sealed` subtypes and pattern matching, I think we actually need to avoid that--because adding a new subtype would break existing code
> 
> Well there's a private `_UnknownType` subclass, similarly to the unknown type in the union enum in generated code. So as long as we're not removing types (which would be breaking with the generated code as well), I don't see how this representation is harder to evolve.

Ah, I missed the `_UnknownType` part, perhaps that is sufficient, thanks.

I'm still wary of introducing too much on top of the raw (extension) types. It's perfectly fine to experiment with though. One other consideration is how CPU time is needed to instantiate everything--since the extension types are "free". There can however be a tradeoff in the other direction where once instantiated the not-extension types are more performant.

> Maybe I'm lacking vision wrt. to the one `Model` approach - if you have ideas on how we'd solve the problem of introducing new types derived from the model without being part of it, I can explore these approaches too.

One idea above :)

I guess the overall vision can be described as "build in the simplest way possible but no simpler", we are still figuring out exactly where that boundary is :) ... I guess important considerations are:

 - ease of use for macro authors, which tends to push to more complexity
 - maintainability (for the SDK / dart_model) as the language changes, which I think most pushes for simplicity
 - performance, which might push in either direction :)

simolus3:
> I'm still wary of introducing too much on top of the raw (extension) types. It's perfectly fine to experiment with though

That's fair - I've kept them for now and implemented the analyzer part of the type hierarchy in the model. The translation is pretty straightforward with the current representation.

> since the extension types are "free".

Re performance, my intuition is that this will depend on the types are used - especially when doing lots of subtype checks the representation based on classes should be much faster since we don't have to extract the shape of the type from the map every time.

At the moment, the API usage for macro authors is this:

```dart
final typeSystem = StaticTypeSystem(model.types);
// Direct schema representation from the type hierarchy
final rawTargetClass = StaticTypeDesc.namedTypeDesc(model.types.named[request.target.name]!.self);
// Converted to resolved representation for type system
final targetClass = StaticType(rawTargetClass);
assert(
  // Assert that the resolved class type we're running against is a subtype
  // of Object.
  typeSystem.isSubtype(
    targetClass,
    InterfaceType(name: InterfaceType.objectName, instantiation: const []),
  ),
);
```

The syntactic conversion overhead is not great, but I think with a `Model` in the `Zone` it could look like this:

```dart
final targetType = modelType(request.target); // directly return resolved representation
assert(typeSystem().isSubtype(targetClass, InterfaceType(...)))
```

Does that roughly match your expectations on how types should work in macros? If the general design looks acceptable, I'll finish with more tests and the CFE implementation.

davidmorgan:
> > I'm still wary of introducing too much on top of the raw (extension) types. It's perfectly fine to experiment with though
> 
> That's fair - I've kept them for now and implemented the analyzer part of the type hierarchy in the model. The translation is pretty straightforward with the current representation.
> 
> > since the extension types are "free".
> 
> Re performance, my intuition is that this will depend on the types are used - especially when doing lots of subtype checks the representation based on classes should be much faster since we don't have to extract the shape of the type from the map every time.

Yes, could be.

> At the moment, the API usage for macro authors is this:
> 
> ```dart
> final typeSystem = StaticTypeSystem(model.types);
> // Direct schema representation from the type hierarchy
> final rawTargetClass = StaticTypeDesc.namedTypeDesc(model.types.named[request.target.name]!.self);
> // Converted to resolved representation for type system
> final targetClass = StaticType(rawTargetClass);
> assert(
>   // Assert that the resolved class type we're running against is a subtype
>   // of Object.
>   typeSystem.isSubtype(
>     targetClass,
>     // We can create new type instances in user code:
>     InterfaceType(name: InterfaceType.objectName, instantiation: const []),
>   ),
> );
> ```
> 
> The syntactic conversion overhead is not great, but I think with a `Model` in the `Zone` it could look like this:
> 
> ```dart
> final targetType = modelType(request.target); // directly return resolved representation
> assert(typeSystem().isSubtype(targetClass, InterfaceType(...)))
> ```
> 
> Does that roughly match your expectations on how types should work in macros? If the general design looks acceptable, I'll finish with more tests and the CFE implementation.

Much like that, yes; but since we can get to `modelType` and `typeSystem` from anywhere we could also do that inside extension methods, so:

```
request.target.isSubtype(InterfaceType(...))
```

should also be possible?

I didn't look in detail at the code yet, please let me know when you think it's ready for careful review / which parts, happy to take a look.

I'm not quite sure what you mean by "CFE implementation"?

simolus3:
I agree with the `isSubtype` extension, it's very neat!

> I didn't look in detail at the code yet, please let me know when you think it's ready for careful review / which parts, happy to take a look.

To get the representation exactly right we'll need a way to represent nullable fields in the model, but I think the structure of how this would represent types and the type hierarchy is already there, so getting feedback on the schema changes would be helpful.
Then `dart_model/lib/src/type.dart` contains the wrappers introduced to represent types without requiring an additional context of which type parameters are in scope. The rest builds ontop of these wrappers, so if anything stands out in there that would be helpful feedback too.

Then finally the subtype check in `src/type_system.dart` is also implemented, but it's a pretty boring translation of the spec so I think only perhaps the parts of how types are looked up from the hierarchy (`_lookupNamed` and `_constructSuperTypes`) are worth looking at for an initial review.

> I'm not quite sure what you mean by "CFE implementation"?

I meant to say that I've added support for including the type hierarchy in `_analyzer_macros` already, but not in `_cfe_macros`.

davidmorgan:
I took a careful look today, this all looks good to me: questions about the exact best API to present, performance and so on I think we can best answer by landing this and iterating, particularly as we build out queries and some first example macros.

If you'd prefer to land without a CFE implementation yet then I suggest a small tweak to the golden files test so it notices if there is e.g. `.json.analyzer` and `.json.cfe` for cases where there are two different goldens.

Thanks :)

simolus3:
> If you'd prefer to land without a CFE implementation yet then I suggest a small tweak to the golden files test so it notices if there is e.g. `.json.analyzer` and `.json.cfe` for cases where there are two different goldens.

I've added a CFE implementation as well. I'm not too familiar with CFE internals so there might be some more efficient ways, but it appears to work with the golden file (the only difference is that the analyzer has a way to represent type parameters without a bound, in Kernel there's an implicit `extends Object?`). So the goldens are slightly different for now.

davidmorgan (approved):
Looks good to me, thanks!

davidmorgan:
Looking forward to experimenting with this, thanks.

FYI

I'm working on a few changes to how the extension types generation works: first to switch from "read schema, generate code" to defining the types separately (in Dart) and generating both schema and code; this should make it a bit easier to extend. Then I'm working on integrating the new JsonBufferBuilder end to end, which involves introducing the idea of a "current model" discussed on this PR. And finally there are a few pending missing features that should be added to the JSON generation: "enums", and from this PR I noticed it would be nice if we have a way to add custom factory constructors / methods.

I mention in case you were planning on extending the JSON generation for better nullability support, if so please wait as that change will certainly clash with mine :)

davidmorgan:
Now merged :) https://github.com/dart-lang/macros/pull/55

PR #41, opened 2024-08-13
Add golden tests for introspection.

davidmorgan:
I guess golden tests will be a good way for us to get introspection working as we want and also make sure the CFE and analyzer match.

This is a rough PR to get something in place:

 - No CFE test yet, but it will be added next, it will be very similar and read the same source + JSON as input, that's why the files are pulled out to a top level `introspection_goldens` folder rather than next to the analyzer tests; I guess we might also need a "skip list" for golden tests not working on one frontend yet
 - It's a bit awkward that the source has to explicitly add `QueryMacro` to everything; instead of doing introspection exactly via a macro we could try to do it with just the "query service", but it worries me a little that that would not exactly match what a macro does, so even though it's a bit awkward I think this makes sense
 - I guess we need to support having a different golden per phase

Any feedback+suggestion welcome :)

### DIFF
--- a/pkgs/_analyzer_macros/test/golden_test.dart
+++ b/pkgs/_analyzer_macros/test/golden_test.dart
@@ -1,0 +1,107 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+import 'dart:isolate';
+
+import 'package:_analyzer_macros/macro_implementation.dart';
+import 'package:analyzer/dart/analysis/analysis_context.dart';
+import 'package:analyzer/dart/analysis/analysis_context_collection.dart';
+import 'package:analyzer/dart/analysis/results.dart';
+import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+void main() {
+  late AnalysisContext analysisContext;
+
+  group('analyzer with injected macro impl query result matches golden', () {
+    final directory = Directory(
+        Isolate.resolvePackageUriSync(Uri.parse('package:foo/foo.dart'))!
+            .resolve('../../../introspection_goldens')
+            .toFilePath());
+
+    setUp(() async {
+      // Set up analyzer.
+      final contextCollection =
+          AnalysisContextCollection(includedPaths: [directory.path]);
+      analysisContext = contextCollection.contexts.first;
+      injected.macroImplementation = await AnalyzerMacroImplementation.start(
+          packageConfig: Isolate.packageConfigSync!);
+    });
+
+    for (final file in directory
+        .listSync(recursive: true)
+        .whereType<File>()
+        .where((f) => f.path.endsWith('.dart'))) {
+      final path = file.path;
+      final relativePath = p.relative(path, from: directory.path);
+
+      path.substring(path.indexOf('introspection_goldens') +
+          'introspection_goldens'.length);
+
+      late File goldenFile;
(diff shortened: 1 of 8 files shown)


davidmorgan:
Test is failing on Windows, will fix but it might take a little while if I need my windows cloudtop.

pkgs/_analyzer_macros/test/golden_test.dart
```diff
@@ -0,0 +1,81 @@
...
+import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
+import 'package:test/test.dart';
+
+void main() {
+  late AnalysisContext analysisContext;
+
+  group('analyzer with injected macro impl query result matches golden', () {
+    setUp(() async {
+      // Set up analyzer.
+      final directory = Directory.current.absolute.uri
```
jakemac53, line 22: Fwiw, a slightly  more robust way of doing this would be something like this:

`Isolate.resolvePackageUriSync(Uri.parse('package:foo/some_file.txt').resolve('../../'))`.

Specifically, it would allow running `dart golden_test.dart` from any directory which might be nice.

pkgs/_analyzer_macros/test/golden_test.dart
```diff
@@ -0,0 +1,81 @@
...
+          .resolve('../../introspection_goldens/foo');
+      final contextCollection =
+          AnalysisContextCollection(includedPaths: [directory.toFilePath()]);
+      analysisContext = contextCollection.contexts.first;
+      injected.macroImplementation = await AnalyzerMacroImplementation.start(
+          packageConfig: Isolate.packageConfigSync!);
+    });
+
+    for (final file in Directory.fromUri(
```
jakemac53, line 31: nit: assign this directory logic to a top level variable, its a fair bit to repeat

pkgs/_analyzer_macros/test/golden_test.dart
```diff
@@ -0,0 +1,81 @@
...
+            Directory.current.uri.resolve('../../introspection_goldens'))
+        .listSync(recursive: true)
+        .whereType<File>()
+        .where((f) => f.path.endsWith('.dart'))) {
+      final path = file.path;
+      final relativePath = path.substring(
```
jakemac53, line 37: nit: let's import and use package:path? I know this isn't super complicated but it also has a more robust implementation.

pkgs/_analyzer_macros/test/golden_test.dart
```diff
@@ -0,0 +1,81 @@
...
+          path.indexOf('introspection_goldens') +
+              'introspection_goldens'.length);
+
+      test(relativePath, () async {
+        final golden = json
+            .decode(File(path.replaceAll('.dart', '.json')).readAsStringSync());
```
jakemac53, line 43: I would use package:path to replace the extension as well

jakemac53 (commented):
This generally LGTM but I would like a workflow for automatically updating the goldens. Goldens which can't be auto updated are super painful.

jakemac53 (approved):

pkgs/_analyzer_macros/test/golden_test.dart
davidmorgan: Sounds good to me :) done, thanks.

pkgs/_analyzer_macros/test/golden_test.dart
davidmorgan: Done.

pkgs/_analyzer_macros/test/golden_test.dart
davidmorgan: Done.

It occurred to me yesterday that it would be really nice if we had lints around path handling, I never know quite what to reach for.

davidmorgan (commented):
Thanks! Done and added golden update workflow. I'll send the equivalent for the CFE and address any comments on that, in both :)

PR #43, opened 2024-08-14
Allow union types to have additional fields, support required fields.

jakemac53:
- Union types can now have additional fields beyond just `type` and `value`. These can be passed to each specialized constructor along with the value.
- Adds support for required fields in schemas, which essentially just makes the parameters for these fields in the constructors be required. These are not intended to be widely used, but there are certain things which are truly required.
- Uses these features to add a required `id` field to request types, and `requestId` field to response types.
- These will be used to support multiple ongoing requests and remove hacks with global variables for cfe/analyzer internals in the future.
- Also dropped the special `_dartWrapperToJson` function - afaik this was never necessary (but please let me know if that is wrong).

If you would like, I could separate out the support here from adding the actual required fields.

## Risks

This makes requests non-hermetic, because of the `id` field. This is always only in the _outer_ request type though. We can still do caching based on the _inner_ request type, which I think should work out fine.

Note that having the IDs on the outer types means they are not generally visible to the user APIs which is good.

### DIFF
--- a/pkgs/macro_service/lib/src/macro_service.dart
+++ b/pkgs/macro_service/lib/src/macro_service.dart
@@ -28,7 +28,7 @@ abstract interface class MacroService {
 /// and that is allowed.
 int get nextRequestId {
   final next = _nextRequestId++;
-  if (_nextRequestId > 2 ^ 32) {
+  if (_nextRequestId > 0x7fffffff) {
     _nextRequestId = 0;
   }
   return next;
--- a/pkgs/macro_service/lib/src/macro_service.dart
+++ b/pkgs/macro_service/lib/src/macro_service.dart
@@ -18,3 +18,20 @@ abstract interface class MacroService {
   /// Handles [request].
   Future<Response> handle(HostRequest request);
 }
+
+/// Shared implementation of auto incrementing 32 bit IDs.
+///
+/// These roll back to 0 once it is greater than 2^32.
+///
+/// These are only unique to the process which generates the request,
+/// so for instance the host and macro services may generate conflicting ids
+/// and that is allowed.
+int get nextRequestId {
+  final next = _nextRequestId++;
+  if (_nextRequestId > 2 ^ 32) {
+    _nextRequestId = 0;
+  }
+  return next;
+}
+
+int _nextRequestId = 0;
--- a/schemas/macro_service.schema.json
+++ b/schemas/macro_service.schema.json
@@ -62,6 +62,10 @@
     "HostRequest": {
       "$description": "A request sent from host to macro.",
       "properties": {
+        "id": {
+          "description": "The id of this request, must be returned in responses.",
+          "type": "integer"
+        },
         "type": {
           "type": "string"
         },
@@ -72,7 +76,12 @@
             }
           ]
         }
-      }
+      },
+      "required": [
+        "id",
+        "type",
+        "value"
+      ]
     },
     "MacroDescription": {
       "type": "object",
@@ -103,6 +112,10 @@
     "MacroRequest": {
       "$description": "A request sent from macro to host.",
       "properties": {
+        "id": {
+          "description": "The id of this request, must be returned in responses.",
+          "type": "integer"
+        },
         "type": {
           "type": "string"
         },
@@ -116,7 +129,12 @@
             }
           ]
         }
-      }
+      },
+      "required": [
+        "id",
+        "type",
+        "value"
+      ]
     },
     "QueryRequest": {
       "type": "object",
@@ -139,6 +157,10 @@
     "Response": {
       "$description": "A response to a [MacroRequest] or [HostRequest].",
       "properties": {
+        "requestId": {
+          "description": "The id of the [AugmentRequest] this is responding to.",
+          "type": "integer"
+        },
         "type": {
           "type": "string"
         },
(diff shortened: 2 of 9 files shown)


pkgs/macro_service/lib/src/macro_service.dart
```diff
@@ -18,3 +18,20 @@ abstract interface class MacroService {
...
+/// Shared implementation of auto incrementing 32 bit IDs.
+///
+/// These roll back to 0 once it is greater than 2^32.
+///
+/// These are only unique to the process which generates the request,
+/// so for instance the host and macro services may generate conflicting ids
+/// and that is allowed.
+int get nextRequestId {
+  final next = _nextRequestId++;
+  if (_nextRequestId > 2 ^ 32) {
```
davidmorgan, line 31: Ehhm ... 2 ^ 32 is 34 . Maybe 0x7fff ffff?

schemas/macro_service.schema.json
```diff
@@ -139,6 +157,10 @@
     "Response": {
       "$description": "A response to a [MacroRequest] or [HostRequest].",
       "properties": {
+        "requestId": {
```
davidmorgan, line 160: FWIW I would also be okay with just "id", but possibly this is a bit clearer, up to you :)

schemas/macro_service.schema.json
```diff
@@ -139,6 +157,10 @@
...
+          "description": "The id of the [AugmentRequest] this is responding to.",
```
davidmorgan, line 161: I think that should be "id of the [MacroRequest] or [HostRequest]..."?

davidmorgan (approved):
Nice, thanks, I was thinking we'd need something along these lines :)

A couple of high level thoughts for later, not blocking for this PR:

1) A slightly different way to extend the union types would be to switch from union types to union fields, so a normal object type can have zero, one or more `oneOf` properties inside it, each of which gets its own type property next to it, generated enum, and pile of `is`/`as` methods. I can't think of any obvious reason why we want this, but it's worth bearing in mind in case it looks neater at some point :)

2) All the repetition in the schema is getting annoying, I feel like we are getting gradually closer to wanting to generate the schema too. For example we could flip the generator around, describe the schema in Dart code instead of in JSON and write out both Dart data types and schema. What do you think? If we do that we should also do the TODO about being able to generate and test with a strict schema, at the same time.

pkgs/macro_service/lib/src/macro_service.dart
jakemac53: lmao!

schemas/macro_service.schema.json
jakemac53: I think this is better - harder to make a mistake when creating a response object

schemas/macro_service.schema.json
jakemac53: I just changed it to the generic "request" since  that will not get out of date later

jakemac53:
> 2\. All the repetition in the schema is getting annoying, I feel like we are getting gradually closer to wanting to generate the schema too. For example we could flip the generator around, describe the schema in Dart code instead of in JSON and write out both Dart data types and schema. What do you think? If we do that we should also do the TODO about being able to generate and test with a strict schema, at the same time.

I am fine with the schema for now. It would be nice to be able to write in a bit more of an OOP style with interfaces and such, to remove duplication, but I think it might be over complicating things. The current approach is alright.

It is also hard because extension types can't implement interfaces (other than the representation type), so I am not sure how we would describe the data types and also extend their implementations with all the Dart boilerplate.

davidmorgan:
> > 2. All the repetition in the schema is getting annoying, I feel like we are getting gradually closer to wanting to generate the schema too. For example we could flip the generator around, describe the schema in Dart code instead of in JSON and write out both Dart data types and schema. What do you think? If we do that we should also do the TODO about being able to generate and test with a strict schema, at the same time.
> 
> I am fine with the schema for now. It would be nice to be able to write in a bit more of an OOP style with interfaces and such, to remove duplication, but I think it might be over complicating things. The current approach is alright.
> 
> It is also hard because extension types can't implement interfaces (other than the representation type), so I am not sure how we would describe the data types and also extend their implementations with all the Dart boilerplate.

Not sure if I was clear enough to actually convey what I meant :) by "describe the schema in Dart code" I didn't mean real Dart code, which would definitely complicate things, I meant some DSL that we invent to be essentially the JSON schema minus the repetition.

Anyway, no rush, if we need it it will be obvious :) the other thing we may end up doing is adding validation, so you can't write something close to a valid union type but not actually valid, but that seems like more work and possibly less good.

Thanks.

jakemac53:
> Not sure if I was clear enough to actually convey what I meant :) by "describe the schema in Dart code" I didn't mean real Dart code, which would definitely complicate things, I meant some DSL that we invent to be essentially the JSON schema minus the repetition.

Ah ok, we could do that too, but I think it sounds like a lot of work for not a lot of benefit. Any time we want to support something new (for instance, required fields like in this PR), we have to update our special DSL parser etc where I mostly got that for free here.

davidmorgan:
By "some DSL" I meant the super low effort type of DSL where you just instantiate your data description :) so basically remove the JSON schema parsing and instantiate the data directly.

```
  generateDartAndSchema([
    JsonClass('Foo', {'foo': JsonType.int, 'bar': JsonType.string}),
    JsonClass('FooBar', {'foo': JsonType.string, JsonType.ref('Foo')}),
    JsonUnion('Baz', [JsonType.string, JsonType.ref('Foo')]),
  ])
```

Agreed we probably don't want to invent a new type of thing to parse.

Anyway, not urgent.

PR #44, opened 2024-08-15
Support binary serialization.

davidmorgan:
Introduce new type `Protocol` for wire format and version; for now it's just set explicitly everywhere, it should be negotiated.

Move serialization logic to this type, including new code to split a binary stream into messages using length-then-message encoding.

Add support for lists to `JsonBuffer`.

Interesting things after this:

 - Add some way to ensure we are using `LazyMap` instead of `Map` where it matters
 - Protocol versioning, negotiation

### DIFF
--- a/pkgs/macro_service/lib/src/macro_service.dart
+++ b/pkgs/macro_service/lib/src/macro_service.dart
@@ -2,7 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:dart_model/serialization.dart';
+
 import 'macro_service.g.dart';
+import 'message_grouper.dart';
 
 /// Service provided by the host to the macro.
 abstract interface class HostService {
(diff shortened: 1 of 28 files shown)


jakemac53:
> * Add some way to ensure we are using `LazyMap` instead of `Map` where it matters

I was thinking of doing this in the generator so if you use the regular extensions type constructors instead of calling `toJson` with a map it would call it with a lazy map.

pkgs/macro_service/lib/src/macro_service.dart
```diff
@@ -18,3 +25,96 @@ abstract interface class MacroService {
...
+      throw StateError('Unsupported protocol: $this');
+    }
+  }
+}
+
+/// Splits the binary format to messages and deserializes them.
+///
+/// Messages are a four byte little-endian length followed by the message data
+/// of that length.
+class _BinaryDeserializer
```
jakemac53, line 78: See https://github.com/dart-lang/sdk/blob/main/pkg/_macros/lib/src/executor/message_grouper.dart. I would just copy that implementation.

We still do end up copying bytes but it uses all Uint8Lists allocated with the correct size and sets ranges within that which is a bit more optimized.

To avoid a copy we would need a buffer which can read from a list of Uint8List (views), which we could also do to be sure. But, then it won't be a Uint8List and as a result might miss out on some other optimizations, so we would need to benchmark.

pkgs/macro_service/lib/src/macro_service.g.dart
```diff
@@ -157,6 +157,18 @@ extension type MacroRequest.fromJson(Map<String, Object?> node) {
   }
 }
 
+/// The macro to host protocol version and encoding. TODO(davidmorgan): add the version.
+extension type Protocol.fromJson(Map<String, Object?> node) {
+  Protocol({
+    String? encoding,
```
jakemac53, line 163: We should make the encoding be an enum (can still be represented as a String on the wire). But the Dart code shouldn't be using strings to create protocols. Might need to update the generator for that but I think it is worth doing.

If we want to leave the door open to more protocols later, we should do the private enum value trick to prevent exhaustive matching.

Could be a TODO though.

pkgs/macro_service/lib/src/macro_service.dart
davidmorgan: Whoops, should have realized there would be an implementation already. Will swap it in + address other comments tomorrow. Thanks :)

davidmorgan:
Thanks, PTAL :)

Re: LazyMap, thanks for the PR on that, will comment there :)

jakemac53 (approved):

PR #45, opened 2024-08-15
use LazyMap in extension constructors

jakemac53:
We don't have a benchmark and might want to wait to land this until we do. Thoughts?

Mostly just wanted to hack on something for the last bit of my day 🤣

### DIFF
--- a/pkgs/dart_model/lib/src/dart_model.g.dart
+++ b/pkgs/dart_model/lib/src/dart_model.g.dart
@@ -1,13 +1,20 @@
 // This file is generated. To make changes edit schemas/*.schema.json
 // then run from the repo root: dart tool/dart_model_generator/bin/main.dart
 
+import 'json_buffer.dart' show LazyMap;
+
 /// An augmentation to Dart code. TODO(davidmorgan): this is a placeholder.
 extension type Augmentation.fromJson(Map<String, Object?> node) {
   Augmentation({
     String? code,
-  }) : this.fromJson({
-          if (code != null) 'code': code,
-        });
+  }) : this.fromJson(LazyMap(
+            [
+              if (code != null) 'code',
+            ],
+            (key) => switch (key) {
+                  'code' => code,
+                  _ => null,
+                }));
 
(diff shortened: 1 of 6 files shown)


pkgs/dart_model/lib/src/dart_model.g.dart
```diff
@@ -1,13 +1,20 @@
...
+
 /// An augmentation to Dart code. TODO(davidmorgan): this is a placeholder.
 extension type Augmentation.fromJson(Map<String, Object?> node) {
   Augmentation({
     String? code,
-  }) : this.fromJson({
-          if (code != null) 'code': code,
-        });
+  }) : this.fromJson(LazyMap(
+            [
```
jakemac53, line 11: Fwiw, I don't love that we will allocate a list still for each of these. It might be worth trying to some up with something better?

rakudrama: The code also allocates a closure over `this`.
The per-object closure could be fixed by making `this` and explicit argument to the closure, and using a static method.

Calling the closure for each present key is likely to be O(N^2) as the switch-on-string generally compiles to an if-then-else chain on the string comparisons. For small N, no problem, but at dome point all the string comparisons will dominate,

davidmorgan: Thanks! I was thinking along these lines yesterday, it was great to find an implementation in my inbox this morning, I spent today adding some benchmarking on top and exploring a third way, writeup here

https://github.com/dart-lang/macros/issues/48

tl;dr this is already nearly 3x faster than using standard maps; I think we can get rid of the list+closure using extension type builders, and it does give some further speedup, to 3.5x; still some way from proving that it works though, input and suggestions welcome :)

davidmorgan:
I think this is obsoleted by https://github.com/dart-lang/macros/pull/51

It's nevertheless useful to benchmark with, I based one benchmark in https://github.com/dart-lang/macros/pull/52 on the generated code here.

PR #47, opened 2024-08-16
Experiment with ways of getting data into JsonBuffer.

davidmorgan:


pkgs/dart_model/benchmark/json_buffer.dart
```diff
@@ -3,20 +3,88 @@
...
+Uint8List? serialized0;
 Uint8List? serialized;
+Uint8List? serialized2;
 JsonBuffer? deserialized;
-const mapSize = 1000000;
+const mapSize = 10000;
 final mapKeys = List.generate(mapSize, (i) => i.toString());
 
 void main() {
+  JsonBufferSerializeBenchmark0().report();
```
jakemac53, line 19: Fwiw calling each benchmark 3 times shouldn't be necessary... it does its own warmup. It is possible the warmup isn't running enough times or something? We can override how warmup works if we need which might be better?

pkgs/dart_model/benchmark/json_buffer.dart
```diff
@@ -25,7 +93,75 @@ class JsonBufferSerializeBenchmark extends BenchmarkBase {
 
   @override
   void run() {
-    serialized = JsonBuffer(LazyMap(mapKeys, (key) => key)).serialize();
+    serialized = JsonBuffer(LazyMap(mapKeys, (key) {
```
jakemac53, line 96: Using a single canonical `mapKeys` list like this is cheating some, I don't think we will ever be able to use pre-existing lists for keys in practice because we are always providing a filtered view on objects (except maybe if somebody asks for _every_ field we could special case that).

Unless we always serialize null values too, but if we were going to do that I would probably advocate for an entirely different approach to serialization :).

pkgs/dart_model/benchmark/json_buffer.dart
```diff
@@ -39,3 +175,191 @@ class JsonBufferDeserializeBenchmark extends BenchmarkBase {
...
+  /// Map of members by name.
+  Map<String, Member> get members => (node['members'] as Map).cast();
+
+  /// The properties of this interface.
+  Properties get properties => node['properties'] as Properties;
+}
+
+extension type InterfaceBuilder._(JsonBuffer buffer) {
+  factory InterfaceBuilder() {
+    runningBuffer!.startMap(2);
```
jakemac53, line 212: I think if we were going to do an approach like this, we would want something like a `JsonBufferView` which could encapsulate the logic of the reserved space for the map, with the ability to allocate new objects in the original buffer and overwrite the original pointer values to point at those, so that we could hopefully remove this entire notion of `runningBuffer`.

Yes, it would mean additional allocations but this approach seems too brittle to me.

pkgs/dart_model/benchmark/json_buffer.dart
davidmorgan: I added it to get an idea of the variance, then noticed that indeed the first run of the first benchmark was slower. Looking closer the runs do not seem to be particularly stable, sometimes there is a first benchmark difference, sometimes not, and there are quite large differences between runs ... I should really be running AOT compiled anyway, I guess, will switch to doing that.

pkgs/dart_model/benchmark/json_buffer.dart
davidmorgan: Yes, will see if this can be a bit more realistic.

pkgs/dart_model/benchmark/json_buffer.dart
davidmorgan: Will experiment :)

I actually like a lot the idea of having access to the `JsonBuffer` that is where we are building up the response to the current request. The idea of "request scope" is heavily used in Guice, it's natural in server or client code to have exactly one of something per request. In Dart I guess we can do that nicely with zones. We can always pass around context explicitly if that turns out to be too brittle.

Separate to that, it occurred to me that we can probably gain a lot by being upfront about the type information that we have.

So for example for `Properties` we know the set of keys and we know the values are all bools.

If we tell `JsonBuffer` what we're doing, it should be possible to write the keys just once, and to pack the bools into bit each. Which will be both smaller and faster :)

davidmorgan (commented):
Thanks! Will see if I can turn this into something actionable without spending too much time on it ;)

pkgs/dart_model/benchmark/json_buffer.dart
davidmorgan: Also, yes, definitely a split to a type for writing and a type for reading.

davidmorgan:
This is obsoleted by https://github.com/dart-lang/macros/pull/51

PR #49, opened 2024-08-16
add 'implements Object' to non-null extension types

jakemac53:
See https://github.com/dart-lang/sdk/issues/56472 for a lengthy discussion on why this is needed.


davidmorgan:
Yeah, this is a weird corner :) LG, thanks.

PR #50, opened 2024-08-22
Make non-required schema properties nullable

simolus3:
At the moment, only a few properties in the macro schemas have been declared as `required`. But even properties that aren't required are exposed as non-nullable getters in the generated code, which is dangerous and also makes properties that can genuinely be null impossible to access safely.

This PR fixes that by:

1. Marking all properties that are expected to be non-null as `required` in the JSON schemas.
2. Supporting default values for primitives that can be used as a fallback (used on the `Properties` class, where values aren't consistently filled out and supposed to be `false` by default I guess?? The current implementation would just throw when accessing fields that haven't been set...)
3. Generating nullable getters for the remaining values.

### DIFF
--- a/pkgs/_analyzer_macros/lib/query_service.dart
+++ b/pkgs/_analyzer_macros/lib/query_service.dart
@@ -31,19 +31,23 @@ class AnalyzerQueryService implements QueryService {
       uris: {
         uri: Library(
           scopes: {
-            clazz.name: Interface(members: {
-              // TODO(davidmorgan): return more than just fields.
-              // TODO(davidmorgan): specify in the query what to return.
-              for (final field in clazz.fields)
-                field.name: Member(
-                    properties: Properties(
-                  isAbstract: field.isAbstract,
-                  isGetter: false,
-                  isField: true,
-                  isMethod: false,
-                  isStatic: field.isStatic,
-                )),
-            })
+            clazz.name: Interface(
+              metadataAnnotations: const [],
+              properties: Properties(),
+              members: {
+                // TODO(davidmorgan): return more than just fields.
+                // TODO(davidmorgan): specify in the query what to return.
+                for (final field in clazz.fields)
+                  field.name: Member(
+                      properties: Properties(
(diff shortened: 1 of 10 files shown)


davidmorgan:
Hi Simon,

Thanks for the PR, however this is not consistent with where we want to take this code, let me explain :)

In anything that might change over time, we can't add required fields: it would break compilation of existing code. Generally speaking "might change over time" includes anything that is modelling the Dart language. So I expect "required" will only be used in a small number of places not related to the language, e.g. it is currently required that requests have an ID.

Throwing on accessing a null is intended behaviour for query responses, which again covers most of the language modelling. We are using "null" to mean "no value was returned" and the default behaviour to throw is because this is a programmer error, you should have requested the data.

For example if we decide `isAbstract` is not something you want by default, then maybe

```
final data = await query.forClass('package:foo/foo.dart#MyClass');
...
for (final member in data.uris['package:foo/foo.dart']!.interfaces['MyClass']
    .members.where((m) => m.isAbstract)) { // throws!
```

needs fixing to

```
final data = await query.forClass('package:foo/foo.dart#MyClass', extraProperties: ['isAbstract]);
...
for (final member in data.uris['package:foo/foo.dart']!.interfaces['MyClass']
    .members.where((m) => m.isAbstract)) {
```

--we expect that when querying macros know what they want, so it will be simple to make the query match the code.

It's definitely true that we might want different semantics for nullability for some cases, if so then we'll need to add them in addition to the currently available semantics.

simolus3:
I see, thanks for taking the time to explain this!

> It's definitely true that we might want different semantics for nullability for some cases, if so then we'll need to add them in addition to the currently available semantics.

Yeah, that was my original motivation for this. E.g. in the static type representation, we may have type parameters with and without a bound (`T` vs. `T extends Object`). These actually need to be nullable, but I think we can represent that in a different way in the schema (and do a runtime distinction based on `data[key]` being null vs. `data` not containing `key`).
I will open another PR for that if that's closer to the semantics we want to have for nullable fields.

PR #51, opened 2024-08-26
Add JsonBufferBuilder.

davidmorgan:
New alternative implementation for `JsonBuffer`, still WIP, will send an email about status.

### DIFF
--- a/pkgs/dart_model/lib/src/json_buffer/closed_map.dart
+++ b/pkgs/dart_model/lib/src/json_buffer/closed_map.dart
@@ -1,0 +1,134 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'json_buffer_builder.dart';
+
+/// Methods for writing and reading "closed maps".
+///
+/// A "closed map" is a `Map<String, Object?>` in a byte buffer that has all
+/// keys and values known at the time of writing.
+extension ClosedMaps on JsonBufferBuilder {
+  // Layout: [length, ...entries]
+  //
+  // Each entry is: [key (pointer to string), value type, value].
+  //
+  // Values are stored with `_writeSmallValueOrPointer`.
+  static const _keySize = _pointerSize;
+  static const _valueSize = _typeSize + _pointerSize;
+  static const _entrySize = _keySize + _valueSize;
+
+  /// Adds a `Map` to the buffer, returns the [Pointer] to it.
+  ///
+  /// The `Map` should be small and already evaluated, making it fast to
+  /// iterate. For large maps see "growable map" methods.
+  Pointer _pointerToClosedMap(Map<String, Object?> map) {
+    explanations?.push('addClosedMap $map');
+
+    final pointer = _nextFree;
+    final length = map.length;
+    _reserve(_lengthSize + length * _entrySize);
+
+    _writeLength(pointer, length);
+
+    var entryPointer = pointer + _pointerSize;
+    for (final entry in map.entries) {
+      final key = entry.key;
+      final value = entry.value;
+      _writePointer(entryPointer, _pointerToString(key));
+      _writeAny(entryPointer + _keySize, value);
+      entryPointer += _entrySize;
+    }
+
+    explanations?.pop();
+    return pointer;
+  }
+
+  /// Returns the [_ClosedMap] at [pointer].
+  Map<String, Object?> readClosedMap(Pointer pointer) {
+    return _ClosedMap(this, pointer);
+  }
+}
+
+class _ClosedMap
+    with MapMixin<String, Object?>, _EntryMapMixin<String, Object?> {
+  final JsonBufferBuilder _buffer;
+  final Pointer _pointer;
+
+  _ClosedMap(this._buffer, this._pointer);
+
+  @override
+  Object? operator [](Object? key) {
+    final iterator = entries.iterator as _ClosedMapEntryIterator;
+    while (iterator.moveNext()) {
+      if (iterator.current.key == key) return iterator.current.value;
+    }
+    return null;
+  }
+
+  @override
+  late final Iterable<String> keys =
+      _ClosedMapEntryIterable(_buffer, _pointer).map((e) => e.key);
+
+  @override
+  late final Iterable<Object?> values =
+      _ClosedMapEntryIterable(_buffer, _pointer).map((e) => e.value);
+
+  @override
+  late Iterable<MapEntry<String, Object?>> entries =
+      _ClosedMapEntryIterable(_buffer, _pointer);
+
+  @override
+  void operator []=(String key, Object? value) {
+    throw UnsupportedError('ClosedMap is readonly.');
+  }
+
+  @override
+  Object? remove(Object? key) {
+    throw UnsupportedError('ClosedMap is readonly.');
+  }
+
+  @override
+  void clear() {
+    throw UnsupportedError('ClosedMap is readonly.');
+  }
+}
+
+/// `Iterable` that reads a "closed map" in a [JsonBufferBuilder].
+class _ClosedMapEntryIterable extends Iterable<MapEntry<String, Object?>> {
+  final JsonBufferBuilder _buffer;
+  final Pointer _pointer;
+
+  _ClosedMapEntryIterable(this._buffer, this._pointer);
+
+  @override
+  Iterator<MapEntry<String, Object?>> get iterator =>
+      _ClosedMapEntryIterator(_buffer, _pointer);
+}
+
+/// `Iterator` that reads a "closed map" in a [JsonBufferBuilder].
+class _ClosedMapEntryIterator implements Iterator<MapEntry<String, Object?>> {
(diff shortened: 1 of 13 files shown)


pkgs/dart_model/lib/src/json_buffer/typed_map.dart
```diff
@@ -0,0 +1,353 @@
...
+  }
+}
+
+class _TypedMap
+    with MapMixin<String, Object?>, _EntryMapMixin
+    implements Map<String, Object?> {
+  final JsonBufferBuilder _buffer;
+  final Pointer _pointer;
+  final Pointer _schemaPointer;
+  late final TypedMapSchema _schema =
```
jakemac53, line 190: nit: make these non-late and add a factory constructor

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
```diff
@@ -0,0 +1,353 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
jakemac53, line 1: I ran out of steam a bit by the time I got to this one admittedly, but I can take a closer look later 🤣

jakemac53:
Definitely liking the general direction here - lots of areas to squeeze out more performance later on too if desired.

I think my main concern overall is we are doing so much work lazily we are hiding a lot of costs and moving them to places that aren't necessarily expected, in the name of "free" deserialization. We should probably be thinking about benchmarks that actually _use_ these deserialized objects in "reasonable" ways, including reading the same fields multiple times, etc.

pkgs/dart_model/lib/src/json_buffer/closed_map.dart
```diff
@@ -0,0 +1,134 @@
...
+  Pointer _pointerToClosedMap(Map<String, Object?> map) {
+    explanations?.push('addClosedMap $map');
+
+    final pointer = _nextFree;
+    final length = map.length;
+    _reserve(_lengthSize + length * _entrySize);
+
+    _writeLength(pointer, length);
+
+    var entryPointer = pointer + _pointerSize;
```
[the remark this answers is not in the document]
davidmorgan: Done.

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  Map<String, Type> toMap() => Map.fromEntries([
+        for (var i = 0; i != _keys.length; ++i)
+          MapEntry(_keys[i], _valueTypes[i]),
+      ]);
+
+  @override
+  bool operator ==(Object other) =>
+      throw UnsupportedError('TypedMapSchema should be compared by identity.');
```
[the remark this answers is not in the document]
davidmorgan: It's because it's easier to test with a deep `hashCode`, but we definitely _don't_ want it for the prod code :) comment added, thanks.

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
davidmorgan: These are intentionally late:

In analyzer+CFE where we are creating extension types "write only", i.e. immediately adding to some other map without keeping a reference, these reads from the buffer are not needed. The type is just serving to pass around the `_pointer` for use as a value by someone else.

Added a comment :)

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  @override
+  void clear() {
+    throw UnsupportedError('JsonBuffer is readonly.');
+  }
+}
+
+/// `Iterable` that reads a `Map` in a [JsonBufferBuilder].
+class _TypedMapEntryIterable
```
davidmorgan: Updated these to specialized for entries / keys / values as per `ClosedMap`.

davidmorgan (commented):
Thanks :)

All replied, and pushed an additional commit instead of breaking everything this time.

You commented that there aren't enough comments on `TypedMap` and I agree, I didn't get around to taking another look at `TypedMap` either, I will do that next :)

Feel free to wait for next commit before taking another look, or to look at the comments+updates I'm sending now, as you like.

pkgs/dart_model/lib/src/json_buffer/explanations.dart
```diff
@@ -0,0 +1,65 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'json_buffer_builder.dart';
+
+/// Global hook for turning on "debug logs" for [JsonBufferBuilder].
+///
+/// Set to an instance of [Explanations] to turn on logging, reset to null to
+/// turn off logging. If set, [JsonBufferBuilder#toString] will print
+/// additional information for each byte.
+Explanations? explanations;
```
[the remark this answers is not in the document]
jakemac53: We could do it with an environment variable so it is const - and then if tests fail you could just set the environment variable to get better output (and maybe we would suggest doing this)?

I don't mind a global if it isn't mutable :)

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -0,0 +1,329 @@
...
+part 'explanations.dart';
+part 'growable_map.dart';
+part 'type.dart';
+part 'typed_map.dart';
+
+class JsonBufferBuilder {
+  /// The root map of the buffer.
+  ///
+  /// It's a "growable map": entries can be added to it.
+  late final Map<String, Object?> map;
```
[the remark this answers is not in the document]
jakemac53: Ah I didn't catch that the methods being called were instance methods, we can keep it this way.

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -220,26 +193,29 @@ class JsonBufferBuilder {
...
-      (_buffer[pointer + 1] << 8) +
-      (_buffer[pointer + 2] << 16) +
-      (_buffer[pointer + 3] << 24);
+  Pointer readPointer(Pointer pointer) => _readUint32(pointer);
 
   /// Reads the uint32 at [Pointer].
-  int readUint32(Pointer pointer) =>
+  int readUint32(Pointer pointer) => _readUint32(pointer);
+
+  int _readUint32(Pointer pointer) =>
```
jakemac53, line 218: nit: Just inline this in `readUint32`?

pkgs/dart_model/lib/src/json_buffer/type.dart
```diff
@@ -0,0 +1,69 @@
...
+  type,
+  pointer,
+  uint32,
+  boolean,
+  stringPointer,
+  closedMapPointer,
+  growableMapPointer,
+  typedMapPointer;
+
+  static Type of(Object? value) {
```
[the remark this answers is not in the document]
jakemac53: Seems fine to me, and if its just for tests then extensions for testing only would also work if we want to not do string comparisons later.

davidmorgan (commented):
Thanks! Addressed comments, and finished cleaning up TypedMap and adding comments.

Elsewhere I tidied up a bit, mostly improving comments, making things private.

PTAL, I think this is now about ready :)

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -297,17 +308,27 @@ class JsonBufferBuilder {
...
     if (!_allowWrites) throw StateError('This JsonBufferBuilder is read-only.');
   }
 
+  /// Explanations of each byte in the buffer.
+  ///
+  /// Run tests with `dart -Ddebug_json_buffer=true test -c source` to enable.
+  ///
+  /// Then, [JsonBufferBuilder#toString] prints additional information for
+  /// each byte.
+  final _Explanations? _explanations =
```
jakemac53, line 322: This should be static yeah?

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -297,17 +308,27 @@ class JsonBufferBuilder {
...
+      const bool.fromEnvironment('debug_json_buffer') ? _Explanations() : null;
```
jakemac53, line 323: I was originally thinking system environment variables not -D ones, but I see now that the former can't be const.

We could file an issue on package:test to support these, fwiw. Not sure we would actually implement it, there could be some pushback, but it might be worth thinking about.

jakemac53 (approved):

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
davidmorgan: Class local is a bit neater for the tests, and it doesn't seem to make a difference in the benchmarks, i.e. performance seems to be the same.

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
davidmorgan: Thanks, filed https://github.com/dart-lang/test/issues/2272

PR #53, opened 2024-08-28
Add closed list, use JsonBufferBuilder instead of JsonBuffer end to end.

davidmorgan:
As expected, straightforward on top of the previous PR :)

With this it works for binary e2e, but still with copying instead and not creating correctly: next update codegen to do that.


jakemac53 (approved):

davidmorgan:
Not sure if there is something I could have done instead of the force push here, because the squashed JsonBufferBuilder PR does not match what was in this PR.

Anyway, no review comments, and it was a pure rebase without changes, so for this case no harm done :) thanks.

PR #57, opened 2024-09-04
Support nullable definitions in schema

simolus3:
This adds support for making some properties nullable in the schema, which is useful to represent things like e.g type parameters which may or may not have an explicit upper bound.
While we technically allow all properties to be absent in the schema to support partial models, nullable properties generate nullable getters in Dart.
An open question is whether we want to distinguish between `null` and "not part of the resolved model" (which we could detect by asserting that the key is present in the map when the getter is invoked).


davidmorgan:
Looks good, thanks!

I'm hoping we don't have to distinguish null from "not passed", as it will be difficult in other places. (Named parameters, binary representation). Doable if we need it, though.

davidmorgan (approved):

PR #58, opened 2024-09-05
Add scope, update generation to use them.

davidmorgan:
WIP


pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -91,6 +81,8 @@ class JsonBufferBuilder {
         return _readUint32(pointer);
       case Type.boolean:
         return _readBoolean(pointer);
+      case Type.anyPointer:
+        return _read(_readType(pointer), pointer + 1);
```
jakemac53, line 85:
```suggestion
        return _read(_readType(pointer), pointer + typeSize);
```

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Runs [function] in this scope.
+  T run<T>(T Function() function) {
+    return runZoned(function, zoneValues: {
+      _symbol: _ScopeData._(this, this == none ? null : JsonBufferBuilder())
+    });
+  }
+
+  /// Runs [function] in this scope.
+  Future<T> runAsync<T>(Future<T> Function() function) async {
```
jakemac53, line 50: Could just return the call to `runZoned` directly here, no need for the async/await afaik?

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -0,0 +1,121 @@
...
+  macro,
+
+  /// Temporary scope for type evaluation.
+  ///
+  /// A new buffer is introduced for temporary intermediate values. It's an
+  /// error to try to serialize it.
+  evaluating;
+
+  /// Runs [function] in this scope.
+  T run<T>(T Function() function) {
```
jakemac53, line 43: Do we want to allow nesting of scopes or should that be an eager error?

jakemac53 (approved):

davidmorgan (commented):
Thanks.

PR #60, opened 2024-09-06
Make QualifiedName a class type with uri and name fields.

davidmorgan:
This seems more maintainable, clearer; and with JsonBufferBuilder it should actually be better for performance: repeated mentions of the same package will get deduped, and no parsing around the `#` is needed.

Added support for adding code from definitions into the generated types, so we can add static methods.

As we had it, this meant QualifiedName needed to be created in a scope. I tried introducing a `host` scope and using the `evaluating` scope more widely to make this work, but didn't get to a state I was happy with: the `host` scope doesn't work well because it would just accumulate temporary `QualifiedName` instances. So instead I ended up simplifying: make buffer-backed types work in `none` scope, get rid of the `evaluating` scope and use `none` instead. If this turns out to be expensive we can of course optimize :) but let's keep it simple for now.


pkgs/dart_model/lib/src/type.dart
```diff
@@ -149,11 +149,11 @@ final class InterfaceType extends StaticType {
     );
   }
 
-  bool get isDartCoreObject => name == objectName;
-  bool get isDartCoreNull => name == nullName;
-  bool get isDartCoreFunction => name == functionName;
-  bool get isDartCoreRecord => name == recordName;
-  bool get isDartAsyncFutureOr => name == futureOrName;
+  bool get isDartCoreObject => name.equals(objectName);
```
jakemac53, line 152: Oof yeah some way to block accidental equality checks would be great here. End users are going to see these extension types...

tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -188,15 +188,22 @@ abstract class Definition {
...
-  /// Defines a class.
+  /// Defines a "class".
+  ///
+  /// Generated as an extension type over `Map<String, Object?>`.
   ///
   /// [createInBuffer] specifies whether the type is written directly to a
   /// buffer when created. If so, it must be created in a `Scope` which will
   /// provide the buffer.
+  ///
+  /// [extraCode] is arbitrary code to write directly in the extension type.
```
jakemac53, line 199: It might be better if we could add stuff like this via mixin or something?

pkgs/dart_model/lib/src/type.dart
davidmorgan: Yes indeed, filed #65.

tool/dart_model_generator/lib/generate_dart_model.dart
davidmorgan: I can't think of any way of getting static methods / constructors ... except augmentations, which we should probably avoid using at least for now.

But very happy to switch to something better if/when it's possible :)

jakemac53: ah, didn't realize you were adding statics

jakemac53 (approved):

PR #61, opened 2024-09-06
Add generation of enums, use an enum for ProtocolEncoding.

davidmorgan:
This seems to work pretty well right up to the point where you try to use `==`, which will likely fail: there is no way to make deserialized strings match the constants.

That seems to be a general problem of extension types: I am thinking we might want a lint that says "never compare or hash extension types". So I guess if we can live with it elsewhere we can live with it for enums.

But, suggestions welcome :)

### DIFF
--- a/pkgs/_analyzer_macros/test/analyzer_test.dart
+++ b/pkgs/_analyzer_macros/test/analyzer_test.dart
@@ -25,7 +25,7 @@ void main() {
           AnalysisContextCollection(includedPaths: [directory.path]);
       analysisContext = contextCollection.contexts.first;
       injected.macroImplementation = await AnalyzerMacroImplementation.start(
-          protocol: Protocol(encoding: 'binary'),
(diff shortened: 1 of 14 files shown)


jakemac53:
> This seems to work pretty well right up to the point where you try to use `==`, which will likely fail: there is no way to make deserialized strings match the constants.

I am a bit confused, it does work in terms of the representation type? Although you have to silence a lint. See https://dartpad.dev/?id=ddd4a1e624db9366e6550ce77531be96

jakemac53:
We could also use actual enums here (where the wire format is still the string name potentially, although it could be an index also).

jakemac53 (approved):

davidmorgan:
Oh, you're right, `==` works fine. Then I thought, surely "identical" doesn't work, but, that works too.

So, yeah, this should actually work pretty well.

Which is good, I wanted to avoid actual enums because that would have a translation/copying cost, and anyway we don't want a sealed union, it should be open. (As now).

Thanks!

PR #63, opened 2024-09-09
Run `dart_model` tests with extra checks enabled.

davidmorgan:
The "-Ddebug_json_buffer" setting also turns on checks for double writes to the same location, which should catch some bugs.

Fix the bug that it does catch :)

The `&& dart test` is a hack, but it seems to work well enough, any other suggestions appreciated :) thanks!

### DIFF
--- a/pkgs/dart_model/mono_pkg.yaml
+++ b/pkgs/dart_model/mono_pkg.yaml
@@ -9,7 +9,7 @@ stages:
     sdk:
     - dev
 - unit_test:
-  - test: --test-randomize-ordering-seed=random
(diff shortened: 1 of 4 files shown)


pkgs/dart_model/mono_pkg.yaml
```diff
@@ -9,7 +9,7 @@ stages:
     sdk:
     - dev
 - unit_test:
-  - test: --test-randomize-ordering-seed=random
+  - test: --test-randomize-ordering-seed=random && dart -Ddebug_json_buffer=true test --test-randomize-ordering-seed=random -c source
```
jakemac53, line 12:
```suggestion
  - command: dart -Ddebug_json_buffer=true test --test-randomize-ordering-seed=random -c source
```

jakemac53 (approved):

PR #64, opened 2024-09-09
Resolve TODOs around multiple macros, when to run macros.

davidmorgan:
This improves the current situation quite a bit, but there is still work to do in this area e.g. shutdown and build again when a macro changes.


pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -96,25 +116,12 @@ class RemoteMacroHost implements Host {
 
   @override
   Future<Model> query(Query query) async {
+    final id = nextRequestId;
```
jakemac53, line 119: Why grab the id early here?

pkgs/_macro_host/lib/macro_host.dart
```diff
@@ -42,47 +42,57 @@ class MacroHost {
...
-    return response.asAugmentResponse;
   }
 }
 
 class _HostService implements HostService {
   final QueryService queryService;
-  // TODO(davidmorgan): this should be per macro, as part of tracking per-macro
-  // lifecycle state.
-  Completer<Set<int>>? _macroPhases;
+  final Map<String, _MacroState> _macroState = {};
```
jakemac53, line 95: Please add a comment describing this, what are the keys?

pkgs/_macro_server/lib/macro_server.dart
```diff
@@ -39,27 +39,73 @@ class MacroServer {
...
+      // Look up the connection with the macro corresponding to `annotation`.
+      connection = _connections
+          .where((c) => c.descriptions.any(
+              (d) => d.annotation.string == request.macroAnnotation.string))
+          .singleOrNull;
+      // If it's found: done.
+      if (connection != null) break;
+      // Not found, wait [timeout] then recheck. Throws `StateError` on
+      // timeout.
+      await _macroDescriptionBecomesKnown.stream.first.timeout(timeout);
```
jakemac53, line 61: How do we know this is the correct one?

pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -48,8 +50,10 @@ class MacroClient {
     return MacroClient._(protocol, macros, socket);
   }
 
-  void _sendRequest(MacroRequest request) {
+  Future<Response> _sendRequest(MacroRequest request) async {
+    final completer = _responseCompleters[request.id] = Completer<Response>();
```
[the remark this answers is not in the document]
davidmorgan: Done.

pkgs/_macro_client/lib/macro_client.dart
```diff
@@ -60,21 +64,37 @@ class MacroClient {
...
     switch (hostRequest.type) {
       case HostRequestType.augmentRequest:
-        await Scope.macro.runAsync(() async => _sendResponse(
-            Response.augmentResponse(
-                await macros.single
-                    .augment(_host, hostRequest.asAugmentRequest),
-                requestId: hostRequest.id)));
+        final macro = macros
+            .where((m) =>
+                m.description.annotation.string ==
```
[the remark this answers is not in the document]
davidmorgan: After #60 this will change to use the "equals" method on `QualifiedName`.

pkgs/_macro_client/lib/macro_client.dart
davidmorgan: Not needed any more, indeed, thanks; done.

pkgs/_macro_host/lib/macro_host.dart
davidmorgan: Done.

pkgs/_macro_server/lib/macro_server.dart
davidmorgan: We don't; execution continues from line 53 which returns if the matching macro is now found or waits again if it's not.

davidmorgan (commented):
Thanks, PTAL.

davidmorgan:
Not sure what I screwed up so that a force push was needed; as far as I can tell nothing surprising got in / the changes are in original commit + address comments commits + merge from main as it should be.

PR #68, opened 2024-09-12
Add a first JSON macro based on package:json/json.dart.

davidmorgan:
Adds a minimal JSON macro, lots of TODOs.

Adds return types for fields to the model + (analyzer only) query handling.

Skip golden tests for analyzer/CFE if there are no goldens, instead of failing.

Uses templating for identifiers as suggested in https://github.com/dart-lang/language/issues/1989#issuecomment-2329438841 ... this is just a proof of concept, we should discuss whether we want templating and if so how it fits into the overall design: it could be client side only or we make the template format part of the protocol. (In the sense of "this string has this specified format which won't change, if you want something different add a new type / field").

I didn't look too closely at the augmentation golden file, just make it approximately match the package:json/json.dart output for the same code. I realize we need yet another type of test, to compile the code with macro applications and then test with it, which in this case would test the JSON produced by the JSON macro. I'll look at how to add that...

### DIFF
--- a/goldens/foo/lib/foo.analyzer.json
+++ b/goldens/foo/lib/foo.analyzer.json
@@ -11,6 +11,16 @@
                 "isField": true,
                 "isMethod": false,
                 "isStatic": false
+              },
+              "returnType": {
+                "type": "NamedTypeDesc",
+                "value": {
+                  "name": {
+                    "uri": "dart:core",
+                    "name": "int"
+                  },
+                  "instantiation": []
+                }
               }
             }
           }
@@ -24,6 +34,16 @@
                 "isField": true,
                 "isMethod": false,
                 "isStatic": false
+              },
+              "returnType": {
+                "type": "NamedTypeDesc",
+                "value": {
+                  "name": {
+                    "uri": "dart:core",
+                    "name": "int"
+                  },
+                  "instantiation": []
+                }
               }
             }
           }
@@ -33,25 +53,6 @@
   },
   "types": {
     "named": {
-      "package:foo/foo.dart#Foo": {
-        "typeParameters": [],
-        "self": {
-          "name": {
-            "uri": "package:foo/foo.dart",
-            "name": "Foo"
-          },
-          "instantiation": []
-        },
-        "supertypes": [
-          {
-            "name": {
-              "uri": "dart:core",
-              "name": "Object"
-            },
-            "instantiation": []
-          }
-        ]
-      },
       "dart:core#Object": {
         "typeParameters": [],
         "self": {
@@ -101,6 +102,25 @@
               }
             }
           ]
+        },
+        "supertypes": [
+          {
+            "name": {
+              "uri": "dart:core",
+              "name": "Object"
+            },
+            "instantiation": []
+          }
+        ]
+      },
+      "package:foo/foo.dart#Foo": {
+        "typeParameters": [],
+        "self": {
+          "name": {
+            "uri": "package:foo/foo.dart",
+            "name": "Foo"
+          },
+          "instantiation": []
         },
         "supertypes": [
           {
--- a/pkgs/_test_macros/lib/json_codable.dart
+++ b/pkgs/_test_macros/lib/json_codable.dart
@@ -25,6 +25,8 @@ class JsonCodableImplementation implements Macro {
   @override
   Future<AugmentResponse> augment(Host host, AugmentRequest request) async {
     return switch (request.phase) {
+      // TODO(davidmorgan): these should be phases 2 and 3, but that doesn't
+      // work right now, it gives no output for phase 3. Investigate and fix.
       1 => phase1(host, request),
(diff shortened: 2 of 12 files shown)


davidmorgan:
Filed an issue #70 for the templating discussion, with some initial thoughts on the topic :)

goldens/foo/lib/foo.analyzer.json
```diff
@@ -11,6 +11,16 @@
                 "isField": true,
                 "isMethod": false,
                 "isStatic": false
+              },
+              "returnType": {
```
jakemac53, line 15: This model for `properties` I think is going to fall apart here. We really don't want to be modeling all the possible properties for all the possible things with a single type?

For instance, `returnType` is really not what the type of a field is, there is a return type on the getter sure, but not the field (it just has a type).

When you are using this `Properties` class, you are also going to get terrible auto complete, with all possible things that might exist on all possible declaration kinds?

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -0,0 +1,170 @@
...
+
+  Future<AugmentResponse> phase2(Host host, AugmentRequest request) async {
+    final result = <Augmentation>[];
+
+    final target = request.target;
+    final model = await host.query(Query(target: target));
+    final clazz = model.uris[target.uri]!.scopes[target.name]!;
+    final initializers = <String>[];
+    for (final field
+        in clazz.members.entries.where((m) => m.value.properties.isField)) {
```
jakemac53, line 52: Today, is it really easy to emit a reference to a member in code, but with this model the Member doesn't have the info required to do that at all - you have to cobble together a bunch of info about how you reached the member.

Can we reasonably add a `QualifiedName` to each member, and make these just lists instead of maps? And then you can easily emit references to the qualified names in code, given only the `Member`, and they will be guaranteed safe.

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -0,0 +1,170 @@
...
+      final name = field.key;
+      final type = field.value.returnType;
+      serializers.add("json[r'$name'] = ${_convertTypeToJson(name, type)}");
+    }
+
+    // TODO(davidmorgan): helper for augmenting methods.
+    result.add(Augmentation(code: '''
+$_jsonMapType toJson() {
```
jakemac53, line 75: TODO: handle super types etc :)

Looks like this needs to actually create the `json` variable and return it as well.

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -0,0 +1,170 @@
...
+          case 'bool':
+          case 'String':
+          case 'int':
+          case 'double':
+          case 'num':
+            return '$reference as ${namedType.name.code}$orNull';
+          case 'List':
+            return '$nullCheck [for (final item in $reference '
+                'as {{dart:core#List}}<{{dart:core#Object}}?>) '
+                '${_convertTypeFromJson('item', namedType.instantiation.first)}'
```
jakemac53, line 105: Should this really be `.single`? Presumably these are the type arguments? We probably want to rename `instantiation` to `typeArguments` as well.

jakemac53: We probably also need a check that there were any type arguments provided (I believe the existing macro does this).

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -0,0 +1,170 @@
...
+                ']';
+          case 'Set':
+            return '$nullCheck {for (final item in $reference '
+                'as {{dart:core#Set}}<{{dart:core#Object}}?>) '
+                '${_convertTypeFromJson('item', namedType.instantiation.first)}'
+                '}';
+          case 'Map':
+            return '$nullCheck {for (final (:key, :value) in $reference '
```
jakemac53, line 113: Probably should be checking that the first type argument is in fact `{{dart:core#String}}`, and check that there are exactly two type args.

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -0,0 +1,170 @@
...
+        switch (namedType.name.name) {
+          case 'bool':
+          case 'String':
+          case 'int':
+          case 'double':
+          case 'num':
+            return reference;
+          case 'List':
+          case 'Set':
+            return '$nullCheck {for (final item in $reference) '
```
jakemac53, line 148: This should always be a list instead of always a set

davidmorgan (commented):
Thanks!

I realize I added a lot of TODOs rather than making changes in this PR, hopefully that's not too annoying, they all seem fairly big and about equal priority so splitting to many PRs seems like a good way to keep things moving.

jakemac53 (approved):

amrgetment:
Congratulations 🙌

PR #69, opened 2024-09-12
Add handshake to determine wire encoding and protocol.

davidmorgan:
For now the host simply has one encoding+protocol that it wants to use and it will only use that.

The next PR in this area should be able to add support for actual protocol breaking changes.

This is an important thing to get right so it should come with a design doc :)

I don't think anything else is blocked on this so it's relatively low priority.

### DIFF
--- a/tool/dart_model_generator/lib/definitions.dart
+++ b/tool/dart_model_generator/lib/definitions.dart
@@ -6,6 +6,56 @@ import 'generate_dart_model.dart';
 
 final schemas = Schemas([
   Schema(
+    schemaPath: 'handshake.schema.json',
+    codePackage: 'macro_service',
+    codePath: 'src/handshake.g.dart',
+    rootTypes: ['HandshakeRequest', 'HandshakeResponse'],
+    declarations: [
+      Definition.clazz('HandshakeRequest',
+          description: 'Request to pick a protocol.',
+          properties: [
+            Property(
+              'protocols',
+              type: 'List<Protocol>',
+              description: 'Supported protocols.',
+            ),
+          ]),
+      Definition.clazz('HandshakeResponse',
+          description:
+              'The picked protocol, or `null` if no requested protocol '
+              'is supported.',
+          properties: [
+            Property(
+              'protocol',
+              type: 'Protocol',
+              description: 'Supported protocol.',
+              nullable: true,
+            ),
+          ]),
+      Definition.clazz('Protocol',
+          description: 'The macro to host protocol version and encoding. '
+              'TODO(davidmorgan): add the version.',
+          properties: [
+            Property('encoding',
+                type: 'ProtocolEncoding',
+                description: 'The wire format: json or binary.'),
+            Property('version',
+                type: 'ProtocolVersion',
+                description: 'The protocol version, a name and number.'),
+          ],
+          extraCode: '''
+/// The initial protocol for any `host<->macro` connection.
+static Protocol handshakeProtocol = Protocol(
+    encoding: ProtocolEncoding.json, version: ProtocolVersion.handshake);
+'''),
+      Definition.$enum('ProtocolEncoding',
+          description: 'The wire encoding used.', values: ['json', 'binary']),
+      Definition.$enum('ProtocolVersion',
+          description: 'The protocol version.',
+          values: ['handshake', 'macros1']),
+    ],
+  ),
+  Schema(
       schemaPath: 'dart_model.schema.json',
       codePackage: 'dart_model',
       codePath: 'src/dart_model.g.dart',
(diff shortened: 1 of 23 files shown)


tool/dart_model_generator/lib/definitions.dart
```diff
@@ -5,6 +5,56 @@
...
+                description: 'The protocol version, a name and number.'),
+          ],
+          extraCode: '''
+/// The initial protocol for any `host<->macro` connection.
+static Protocol handshakeProtocol = Protocol(
+    encoding: ProtocolEncoding.json, version: ProtocolVersion.handshake);
+'''),
+      Definition.$enum('ProtocolEncoding',
+          description: 'The wire encoding used.', values: ['json', 'binary']),
+      Definition.$enum('ProtocolVersion',
```
jakemac53, line 53: Should this be a semantic version? For example that could help to model new fields being available for some objects the protocol.

davidmorgan: Good question :)

I think we'll need to work through some examples. (I was only half kidding that maybe we should launch v2 and not v1 so that we practice doing a major version change).

My guess for incremental changes is that we want to model them so that the client can figure out what to do from the data in a fine grained way. e.g. whether a field is supported in a result can be determined by whether it gets filled :) or if it was explicitly requested in a query then the host can respond explicitly "I don't support these fields" as part of the response. That way any conditional code in macros can be targeted at the exact feature rather than a version.

If that's the route we take then we might want to hide the minor version from the macro so that it has to be coded against the data instead of the version :)

We might still want to actually _have_ a minor version e.g. for logging. If so then rather than exactly a semantic version I would suggest we split into major version, for breaking changes, and an arbitrary string for minor version that's just for logging.

jakemac53: Ultimately, we really want this to be solved by pub. That is how users expect to have versions solved for them.

If I depend on `package:macros`, and I see a new field on a type, I expect to just be able to set my `package:macros` constraint to `^<latest-version>` and rely on that field being present for any user who can successfully select my package.

We could manage that by setting SDK constraints in package:macros, but it also seems reasonable for the protocol to use a semantic version as an extra check.

jakemac53: We might need to create a doc or something, but suffice to say I disagree here fundamentally. I think there is only one constraint which is important:

- For any non-breaking SDK release, all existing macros should continue to work.

That is consistent with how the entire ecosystem of packages works today, but it is _not_ true of the current package:macros, because currently the SDK supports exactly one version of that package. That is the problem we need to solve (new SDKs working with old macros), but we don't need (or even want I would argue) new macros to work with old SDKs (this will likely result in confusing errors at compile time).

I understand what you are trying to achieve, but I think it creates far more pain for macro authors and users in the long term. Consider if whenever you wanted to use a new feature from a package, you could not constraint yourself to the version of the package that actually supports that feature, and instead had to do some form of feature detection, with fallback logic (or worse, just throwing an error). This would lead to a massive amount of technical debt as the matrix of possible features you might require expands. With semantic versioning, you can completely ignore this, you simply require the version that actually has the feature, and assume it is there, end of story. There is still a very reasonable story for old versions of the package (or SDK in this case), they just get an older version of your macro. That old version still exists on pub, and will work every bit as well as it did previously.

davidmorgan: Definitely room for discussion, then :)

My experience in maintaining `built_value` through Dart 2 -> 3 suggests to me that I want the feature matrix, not the semver.

Codegen/macros are already written in a "feature matrix" style, they are basically a bunch of switch statements over everything that could show up in the code. Usually new language versions don't remove old things, just add new things. So the old code can stay as it is while I add new case handling next to it. As a codegen/macro author it's simply frustrating that the old code can no longer be compiled everywhere as a result.

For `built_value` there was tension between keeping Dart 2 support going and adding Dart 3 support: it felt like I had no good options, the only "correct" option was to branch but I didn't want to do that work, so I delayed Dart 3 support as long as possible.

With analyzer-based codegen the "can no longer be compiled" was because of analyzer API changes, with `dart_model` we effectively have the option to just skip that version check and see how far the code runs.

We _could_ pass that option on to macro authors, I guess by optional use of SDK constraints on their macro Impl package or perhaps something optional in the protocol.

That way they could for example start following strict semver but have the option to support older SDKs by releasing a new version with relaxed constraints, perhaps in response to feedback.

That kind of option would have significantly helped my `built_value` anxiety :)

Thanks.

jakemac53: Ultimately the whole ecosystem is based on semver. If users don't keep up with new versions they are left behind, but can continue working as normal with the features they have.  I would need a really solid argument for exactly why macros are special here to be convinced that they should work totally differently.

For example, for any Dart package, if it wants to use new features from a dependency (package or SDK), it has to drop support for older versions of those things (or maintain a separate branch). That is just how the system works.

Either way though, it seems like being able to signal in the handshake a full semantic version would be a good thing. That doesn't need to require both sides to be on identical versions, or that the server is >= the client version even necessarily, at least at the protocol level. But it gives clients the option to bail early etc.

jakemac53 (approved):

PR #71, opened 2024-09-25
Add `whereType` extension methods for iterables of union extension types.

davidmorgan:
It occurred to me that this might be a useful type of extension method to add, what do you think?

### DIFF
--- a/pkgs/dart_model/lib/src/dart_model.g.dart
+++ b/pkgs/dart_model/lib/src/dart_model.g.dart
@@ -514,6 +514,33 @@ extension type StaticTypeDesc.fromJson(Map<String, Object?> node)
   }
 }
 
+extension IterableOfStaticTypeDescExtension on Iterable<StaticTypeDesc> {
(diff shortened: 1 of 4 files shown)


pkgs/dart_model/lib/src/dart_model.g.dart
```diff
@@ -514,6 +514,33 @@ extension type StaticTypeDesc.fromJson(Map<String, Object?> node)
   }
 }
 
+extension IterableOfStaticTypeDescExtension on Iterable<StaticTypeDesc> {
+  Iterable<DynamicTypeDesc> whereTypeDynamicTypeDesc() =>
```
jakemac53, line 518: I worry a bit about exposing these Iterables, especially with the likely high cost of iteration (and map lookups involved) due to the underlying representation here, combined with the lack of composability of them.

I also don't love the naming scheme haha. Maybe just remove the first `Type`? So, `whereDynamicTypeDesc`, etc.

PR #74, opened 2024-09-30
Add "macro tool" for running code with macros applied

davidmorgan:
Until we integrate the new code into the build change, we need to glue things together ourselves to do e2e testing.

This is some glue :)

Add a CLI tool that can build+run code with macros.

It needs to create augmentation files, patch to include the augmentations, run the code, then clean up.

It would also work to copy the package to run to somewhere else to make the changes, but I didn't do that.

For end to end testing. This is how I found many issues with the json_codable macro, will send those fixes separately :) ... will follow up adding some actual use of this in tests.

Example output

```
dart ~/git/macros/pkgs/_macro_tool/bin/main.dart --workspace=. --packageConfig=../.dart_tool/package_config.json --script=foo/lib/declare_x.dart
Running declare_x.dart with macros.
~~~
Package config: /usr/local/google/home/davidmorgan/git/macros/.dart_tool/package_config.json
Workspace: /usr/local/google/home/davidmorgan/git/macros/goldens
Script: /usr/local/google/home/davidmorgan/git/macros/goldens/foo/lib/declare_x.dart
Macro output (patched to use augment library): /usr/local/google/home/davidmorgan/git/macros/goldens/foo/lib/declare_x.dart.macro_tool_output
Patching to import augmentation: /usr/local/google/home/davidmorgan/git/macros/goldens/foo/lib/declare_x.dart
~~~ running, output follows
3
~~~ done running, cleanup follows
Reverting: /usr/local/google/home/davidmorgan/git/macros/goldens/foo/lib/declare_x.dart
Deleting: /usr/local/google/home/davidmorgan/git/macros/goldens/foo/lib/declare_x.dart.macro_tool_output
```

### DIFF
--- a/pkgs/_macro_tool/bin/main.dart
+++ b/pkgs/_macro_tool/bin/main.dart
@@ -1,0 +1,37 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:_macro_tool/macro_tool.dart';
+import 'package:args/args.dart';
+import 'package:path/path.dart' as p;
+
+final argParser = ArgParser()
+  ..addOption('workspace', help: 'Path to workspace.')
+  ..addOption('packageConfig', help: 'Path to package config.')
+  ..addOption('script', help: 'Path to script.')
+  ..addFlag('skip-cleanup');
+
+Future<void> main(List<String> arguments) async {
+  final args = argParser.parse(arguments);
+  final workspace = args['workspace'] as String?;
+  final packageConfig = args['packageConfig'] as String?;
+  final script = args['script'] as String?;
+
+  if (workspace == null || packageConfig == null || script == null) {
+    print('''
+Runs a Dart script with `dart_model` macros.
+
+${argParser.usage}''');
+    exit(1);
+  }
+
+  final tool = MacroTool(
+      workspacePath: p.canonicalize(workspace),
+      packageConfigPath: p.canonicalize(packageConfig),
+      scriptPath: p.canonicalize(script),
+      skipCleanup: args['skip-cleanup'] as bool);
+  await tool.run();
+}
--- a/pkgs/_macro_tool/lib/macro_tool.dart
+++ b/pkgs/_macro_tool/lib/macro_tool.dart
@@ -1,0 +1,147 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:_analyzer_macros/macro_implementation.dart';
+import 'package:analyzer/dart/analysis/analysis_context_collection.dart';
+import 'package:analyzer/dart/analysis/results.dart';
+import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
+import 'package:macro_service/macro_service.dart';
+import 'package:path/path.dart' as p;
+
+/// Runs a Dart script with `dart_model` macros.
+class MacroTool {
+  String workspacePath;
+  String packageConfigPath;
+  String scriptPath;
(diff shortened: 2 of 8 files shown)


pkgs/_macro_tool/bin/main.dart
```diff
@@ -0,0 +1,37 @@
...
+  ..addOption('script', help: 'Path to script.')
+  ..addFlag('skip-cleanup');
+
+Future<void> main(List<String> arguments) async {
+  final args = argParser.parse(arguments);
+  final workspace = args['workspace'] as String?;
+  final packageConfig = args['packageConfig'] as String?;
+  final script = args['script'] as String?;
+
+  if (workspace == null || packageConfig == null || script == null) {
```
jakemac53, line 23: Can we not just default at least the workspace and packageConfig args? (to `./` and regular package config discovery based on the script path?).

pkgs/_macro_tool/lib/macro_tool.dart
```diff
@@ -0,0 +1,147 @@
...
+  /// Returns whether an augmentation file was written.
+  Future<bool> _augmentUsingAnalyzer() async {
+    final contextCollection =
+        AnalysisContextCollection(includedPaths: [workspacePath]);
+    final analysisContext = contextCollection.contexts.first;
+    injected.macroImplementation = await AnalyzerMacroImplementation.start(
+        protocol: Protocol(
+            encoding: ProtocolEncoding.binary,
+            version: ProtocolVersion.macros1),
+        packageConfig: File(packageConfigPath).uri);
```
jakemac53, line 88:
```suggestion
        packageConfig: Uri.file(packageConfigPath));
```

pkgs/_macro_tool/lib/macro_tool.dart
```diff
@@ -0,0 +1,147 @@
...
+    final augmentationUnits =
+        resolvedLibrary.units.where((u) => u.isMacroAugmentation).toList();
+    if (augmentationUnits.isEmpty) {
+      return false;
+    }
+
+    print('Macro output (patched to use augment library): '
+        '$_augmentationFilePath');
+    File(_augmentationFilePath).writeAsStringSync(augmentationUnits
+        .first.content
```
jakemac53, line 108:
```suggestion
        .single.content
```

pkgs/_macro_tool/lib/macro_tool.dart
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /// The path where the script file will be backed up before modification.
+  String get _backupScriptPath => '$scriptPath.backup';
+
+  /// Adds `import augment` of the augmentation file.
+  ///
+  /// When macros run in the analyzer or CFE this inclusion of the augmentation
+  /// output is automatic, but for `macro_tool` it has to be patched in.
+  void _addImportAugment() {
```
jakemac53, line 129: Instead of overwriting the actual file, which can leave you in a bad state where you have to manually copy it back, why not just add the import to the new file and execute that?

Then, worst case we just don't do the cleanup.

jakemac53 (approved):

pkgs/_macro_tool/bin/main.dart
davidmorgan: I'd prefer to keep it simple for now, explicit should work nicely for e2e tests.

If we end up actually running it by hand we can make it smarter :)

pkgs/_macro_tool/lib/macro_tool.dart
davidmorgan: Done.

pkgs/_macro_tool/lib/macro_tool.dart
davidmorgan: Done.

pkgs/_macro_tool/lib/macro_tool.dart
davidmorgan: Hmm, I don't follow.

I think the file with the macro application needs the import? I don't see how to keep it in just the new file.

davidmorgan (commented):
Thanks!

I didn't understand one of your comments, please clarify and I'll send a follow-up PR as needed.

PR #75, opened 2024-09-30
Fixes to @JsonCodable.

davidmorgan:
Actually trying to compile the output (per my other PR with macro_tool) led to quite a pile of fixes needed, unsurprisingly :)


davidmorgan:
The output was still not quite correct, with a few more fixes it now works when moved to phase 2+3 as it should be :)

jakemac53 (approved):

PR #76, opened 2024-09-30
Move templating to the client side.

davidmorgan:
Pretty sure there's more to add to the `Augmentation` model representation for other TODOs, but this PR just adds the simplest things that allows the templating to move client side: a split into "resolved strings" and identifiers.


pkgs/_analyzer_macros/lib/macro_implementation.dart
```diff
@@ -9,7 +9,7 @@ import 'package:analyzer/src/summary2/macro_declarations.dart' as analyzer;
 import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
 import 'package:dart_model/dart_model.dart';
 import 'package:macro_service/macro_service.dart';
-import 'package:macros/macros.dart';
+import 'package:macros/macros.dart' hide Code;
```
jakemac53, line 12: If we have two `Code` classes we should really rename one.

Or maybe, it would help to always prefix the one that is used. Cause it is very confusing :)

pkgs/_cfe_macros/lib/macro_implementation.dart
```diff
@@ -121,9 +121,29 @@ class CfeRunningMacro implements injected.RunningMacro {
...
+      // application target. Instead, it should be explicit in
+      // `AugmentResponse`.
+      : typeAugmentations = {
+          // TODO(davidmorgan): empty augmentations response breaks the test,
+          // it's not clear why.
+          if (declarations.isNotEmpty)
+            (target as Declaration).identifier: declarations
+        };
+
+  static Future<CfeMacroExecutionResult> expandTemplates(
```
jakemac53, line 138: This isn't really expanding templates any more right? I would rename it.

It seems like really this is translating between the DartModel code objects and package:macros Code objects?

pkgs/_macro_host/test/macro_host_test.dart
```diff
@@ -62,9 +65,11 @@ void main() {
                     target: QualifiedName(
                         uri: 'package:foo/foo.dart', name: 'Foo'))),
             Scope.macro.run(() => AugmentResponse(augmentations: [
-                  Augmentation(
-                      code:
-                          '// {"uris":{"package:foo/foo.dart":{"scopes":{}}}}')
+                  Augmentation(code: [
+                    Code.resolvedCode(ResolvedCode(
+                        code:
+                            '// {"uris":{"package:foo/foo.dart":{"scopes":{}}}}'))
```
jakemac53, line 71: I have... literally no idea what this is doing 🤣

pkgs/_test_macros/lib/declare_x_macro.dart
```diff
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:dart_model/dart_model.dart';
+import 'package:dart_model/templating.dart';
```
jakemac53, line 6: I would move this all the way out to either a testing utility or the JSON macro example itself etc.

I don't think we should be focusing on shipping a templating syntax as part of this, it isn't a requirement at all for shipping things. And long term probably shouldn't exist at all (we really want custom interpolations via tagged strings or similar).

schemas/dart_model.schema.json
```diff
@@ -8,14 +8,40 @@
...
+        "type": {
+          "type": "string"
+        },
+        "value": {
+          "oneOf": [
+            {
+              "$ref": "#/$defs/QualifiedName"
+            },
+            {
+              "$ref": "#/$defs/ResolvedCode"
```
jakemac53, line 35: I don't really understand what "ResolvedCode" means. What is unresolved code?

pkgs/_macro_host/test/macro_host_test.dart
jakemac53: Looking at this with fresh eyes I see it is really just setting up an expected response.

pkgs/_analyzer_macros/lib/macro_implementation.dart
```diff
@@ -9,7 +9,7 @@ import 'package:analyzer/src/summary2/macro_declarations.dart' as analyzer;
 import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
 import 'package:dart_model/dart_model.dart';
 import 'package:macro_service/macro_service.dart';
-import 'package:macros/macros.dart' hide Code;
+import 'package:macros/macros.dart' as injected;
```
jakemac53, line 12: I generally would prefer prefixes named after the package, so just `macros` in this case.

jakemac53 (commented):
Previous comments still stand regarding ResolvedCode and moving the template expansion completely out to the JSON macro itself

davidmorgan (commented):
Whoops, I had pending comments that I failed to publish, that's why you didn't get the review again. My bad.

I dropped them since we already discussed since I wrote those comments, what I'm publishing now is brand new fresh comments, PTAL ;)

pkgs/_analyzer_macros/lib/macro_implementation.dart
jakemac53: The word `injected` I think in this case is also just super confusing to me? I think because the whole situation is so complicated, I don't really understand what is injected into what or how, and so this term is just confusing... at least for `macros` I know its the package:macros stuff?

jakemac53: I also did not realize we were using overlapping prefixes, that is just even more confusing haha. I would call the other one `analyzer_macros` or something.

schemas/dart_model.schema.json
jakemac53: Do we really need both of these kinds of Code at all? Qualified names are effectively resolved right?

I am pretty confused by what exactly these two things together are accomplishing.

davidmorgan:
Notes from our discussion:

`ResolvedCode` -> `Text`. (Or: check if we can have `String` in a union).

jakemac53 (approved):

pkgs/_analyzer_macros/lib/macro_implementation.dart
davidmorgan: Done.

schemas/dart_model.schema.json
davidmorgan: Changed `ResolvedCode` to `String`, with a small tweak to the generator to make that work. Much better now, thanks :)

Issue #73, opened 2024-09-29
Push resolved model into macro scope?


simolus3:
A while back, we've briefly discussed the idea of putting resolved models into a `Zone` to make models globally accessible: https://github.com/dart-lang/macros/pull/37#issuecomment-2283794860.
Given that scopes have since been implemented, I wonder if resolved `Model`s in macros should be added to the respective macro scope, making the model globally accessible. With an accessible model (and derived structures like `StaticTypeSystem` that could then be cached alongside the model), some APIs would be easier to access (e.g. we could have a `isSubtypeOf` extension on type models instead of forcing users to explicitly construct a type system).

If this is still something that sounds worth exploring, I'm happy to make a contribution in this direction. A big question is how we'd deal with macros calling `query` multiple times, but IIRC we're supposed to be able to combine models eventually so we could still keep an internal model that we add information to on each query result?

davidmorgan:
Hi Simon :)

Yes, definitely, that sounds like a good area to explore, I'll be happy to review PRs, whether work in progress experimentation or landable.

The [json_codable macro](https://github.com/dart-lang/macros/blob/main/pkgs/_test_macros/lib/json_codable.dart) has quite a few TODOs around doing checking of types, in case that's useful to refer to; it's based on the experimental release [here](https://github.com/dart-lang/sdk/blob/main/pkg/json/lib/json.dart) that does do the checks.

Yes, I think accumulating the information from multiple queries is the right way to go.

Thanks!

PR #81, opened 2024-10-01
Push resolved model into macro scope

simolus3:
This adds the `MacroScope` class only exposed publicly via zone values. It has a method to add a new `Model` into an accumulated view that we can expand with multiple resolve calls.

By having this information in the zone, we can make derived utilities (like the `StaticTypeSystem`) available globally, allowing things like `isSubtypeOf` to be added as an extension on static types.

Closes https://github.com/dart-lang/macros/issues/73.

### DIFF
--- a/pkgs/dart_model/lib/src/scopes.dart
+++ b/pkgs/dart_model/lib/src/scopes.dart
@@ -8,7 +8,8 @@ import 'dart:typed_data';
 import '../dart_model.dart';
 import 'json_buffer/json_buffer_builder.dart';
 
-/// Scope for accumulating `dart_model` data for efficient transmission.
+/// Scope for accumulating `dart_model` queries for efficient transmission and
+/// responses for use in type queries.
 ///
 /// A scope introduces buffers that can be written to. Some `dart_model` types
 /// are then instantiated directly into the buffer instead of on the heap,
@@ -128,8 +129,8 @@ class _ScopeData {
   _ScopeData._(this.type, this.buffer, [this.macro]);
 }
 
-/// Data attached to [Scope.macro] environments to make resolved queries
-/// publicly accessible.
+/// Data attached to [Scope.macro] scopes to make resolved queries publicly
+/// accessible.
 class MacroScope {
   Model? _accumulatedModel;
 
@@ -144,7 +145,7 @@ class MacroScope {
   MacroScope._();
 
   /// Returns a type system resolving subtype queries for types known to the
-  /// current Dart model.
+  /// model constructed via [addModel].
   StaticTypeSystem get typeSystem {
     return _typeSystem ??= switch (_accumulatedModel) {
       // TODO: Refactor type system so that we can return an empty type system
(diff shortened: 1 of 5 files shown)


pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -118,6 +123,47 @@ enum Scope {
 class _ScopeData {
   final Scope type;
   final JsonBufferBuilder? buffer;
+  final MacroScope? macro;
+
+  _ScopeData._(this.type, this.buffer, [this.macro]);
+}
+
+/// Data attached to [Scope.macro] environments to make resolved queries
```
davidmorgan, line 131: Nitpick: environments->scopes so we always refer to scopes as ... scopes :)

I often catch myself using synonyms in comments because it's better writing style, but then I realize what we really want is to have a small and precise set of nouns in play ;)

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -118,6 +123,47 @@ enum Scope {
...
+    return _typeSystem ??= switch (_accumulatedModel) {
+      // TODO: Refactor type system so that we can return an empty type system
+      // here that doesn't know anything?
+      null => throw StateError('Cannot obtain type system without a model'),
+      var model => StaticTypeSystem(model.types),
+    };
+  }
+
+  void addModel(Model model) {
+    // TODO: Integrate new model into accumulated model instead of replacing it
```
davidmorgan, line 158: I suggest stating the intent, "merge `model` into the scope's accumulated model", in a doc comment, then adding the TODO right next to that clarifying that that is not what it actually does today :)

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -118,6 +123,47 @@ enum Scope {
...
+  /// Since new models added through [addModel] only contribute new aspects of
+  /// the type system without invalidating earlier incomplete views, all queries
+  /// made through earlier type system instances continue to be valid.
+  StaticTypeSystem? _typeSystem;
 
-  _ScopeData._(this.type, this.buffer);
+  MacroScope._();
+
+  /// Returns a type system resolving subtype queries for types known to the
+  /// current Dart model.
```
davidmorgan, line 147: Maybe:

...known to the model added using [addModel].

?

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async';
 import 'dart:typed_data';
 
+import '../dart_model.dart';
 import 'json_buffer/json_buffer_builder.dart';
 
 /// Scope for accumulating `dart_model` data for efficient transmission.
```
davidmorgan, line 11: Maybe expand this, something like: ...`dart_model` queries for efficient transmission, and responses for use in type queries.

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -118,6 +123,47 @@ enum Scope {
...
+  void addModel(Model model) {
+    // TODO: Integrate new model into accumulated model instead of replacing it
+    _accumulatedModel = model;
+    _typeSystem = null;
+  }
+
+  static MacroScope get current {
+    return switch (Scope._currentOrNull) {
+      _ScopeData(:final macro?) => macro,
+      _ => throw StateError('MacroScope.current can only be called in macros.'),
```
davidmorgan, line 166: I suggest matching the wording on line 65, e.g.

MacroScope.current can only be called in macro scope; but, currently running: ${Scope._currentOrNull?.type ?? Scope.none}

davidmorgan (commented):
Looks great, thanks! Just some suggestions for comments.

davidmorgan (approved):
Thanks :)

Issue #80, opened 2024-10-01
Build and run with macros for e2e testing


davidmorgan:
Hi @johnniwinther, I'm running into what I _think_ might be a CFE issue, could you take a look please?

What I'm trying to do is to take the code output by a new `dart_model` macro, dump it to disk, then run it with the CFE. So from the CFE's point of view they are "just" augmentations; the only weird point is that I have to explicitly add the `import augment` that the CFE would add invisibly when running as a macro.

What I think I see is that the CFE-generated augmentation works but the equivalent augmentation without macros doesn't.

All the files below are in [this draft PR](https://github.com/dart-lang/macros/pull/79/commits/639efa2cf3d85219b707dec80e1eb6ecb113c5d5).

So this works (with CFE build at head today):

```
# declare_define_constructor.dart
import 'declare_define_constructor_macro.dart';

@DeclareDefineConstructor()
class A {}

void main() => print(A.x());
```

```
# declare_define_constructor_macro.dart
import 'package:macros/macros.dart';

macro class DeclareDefineConstructor implements ClassDeclarationsMacro, ClassDefinitionMacro {
  const DeclareDefineConstructor();

  Future<void> buildDeclarationsForClass(
      ClassDeclaration clazz, MemberDeclarationBuilder builder) async {
      builder.declareInType(DeclarationCode.fromParts([
        'external ',
        clazz.identifier.name,
        '.x();'
      ]));
  }

  Future<void> buildDefinitionForClass(
      ClassDeclaration clazz, TypeDefinitionBuilder typeBuilder) async {
    final constructors = await typeBuilder.constructorsOf(clazz);
    final x =
        constructors.singleWhere((c) => c.identifier.name == 'x');
    final builder = await typeBuilder.buildConstructor(x.identifier);
    builder.augment();
  }
}
```

```
dart --enable-experiment=macros --enable-experiment=enhanced-parts declare_define_constructor.dart
Instance of 'A'
```

Looking at this code in the debugger I can see the augmentations that get produced:

```
augment library 'file:///usr/local/google/home/davidmorgan/git/macros/scratch/declare_define_constructor.dart';

augment class A {
external A.x();
  augment A.x();
}
```

and then my equivalent `dart_model` macro produces something very similar, and trying to run (which means adding an explicit `import augment`) it I end up with this on disk:

```
# declare_define_constructor.dart
import augment 'declare_define_constructor.dart.macro_tool_output'; // added by macro_tool
[...]

PR #82, opened 2024-10-02
Match `package:json` @JsonCodable test, add to presubmit

davidmorgan:
Copy in exact test from `package:json` @JsonCodable.
    
Fixes to `dart_model` @JsonCodable, workaround for CFE issue #80.
    
Add e2e test script and add it to CI.

Tweak `_macro_tool`: instead of using a backup file, strip lines tagged as added by the tool. This makes repeats runs of the tool stable if you use `--skip-cleanup`, which is handy when iterating.


tool/run_e2e_tests.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
jakemac53, line 1: In general I am not a fan of tests that can't be ran with just `dart test`. Can we make this be a regular Dart test?

davidmorgan: We _could_, but it also kind of already is, which is why I'm hesitant.

I also like `dart test` :) which is why I like that the underlying binary being built here, `json_codable_test.dart`, is using `package:test`, it's pretty much an exact copy of the normal test in `package:json`.

If we turn `run_e2e_tests.sh` into equivalent Dart code that runs with `dart test`, then when it runs `json_codable_test.dart` there will be a `dart test` running a nested `package:test` test, and I think it'll be pretty hard to tell what's going on, because the outer and inner test case messages, failure messages, etc, will be hard to distinguish.

When you're iterating on one of these tests I think what you should be running is effectively `dart test` but via `_macro_tool`, which means you won't be running the shell script, but the one "test case" that's failing.

Soooo maybe this is a good setup? If we end up wanting any more complexity then it definitely makes sense to have it in dart and not sh. There's still at least one more test type left to add, golden error tests, so maybe take another look after that, what do you think?

Thanks.

jakemac53 (approved):

Issue #72, opened 2024-09-26
Consider making path to node readily available in `Model`


davidmorgan:
Per this comment

https://github.com/dart-lang/macros/pull/68/files#r1759229776

it's really convenient for some use cases if a node in the model knows its own location. For example, that would mean a field in a class knows how it can be referred to from anywhere.

One option would be to change the schema to make that information explicit in the model.

But I wonder if we can cheat. When we instantiate a `Map` from a `JsonBufferBuilder`, it's always either the root map or it's in the context of another `Map` or `List` lookup. A `Map` is just a pointer into `JsonBufferBuilder`, I wonder if we could add also a pointer to the parent, and so provide a fast `Model.pathTo(node)` that uses the parent pointers to compute the path from root to the node.

PR #87, opened 2024-10-04
Add Model.qualifiedNameOfMember, add fast way to find the path to JsonBufferBuilder

davidmorgan:
For #72, this seems reasonably convincing that we can get the path to a node cheaply, which means we don't have to add it everywhere explicitly in the schema.

This doesn't add a use case yet, probably there will be some tweaks when there is a first use case; one specific note is that the link to parent could also store the key in the parent, which would avoid the search through parent's keys that currently happens. But that's easy to change later.

### DIFF
--- a/pkgs/dart_model/lib/src/dart_model.dart
+++ b/pkgs/dart_model/lib/src/dart_model.dart
@@ -14,14 +14,19 @@ extension QualifiedNameExtension on QualifiedName {
 }
 
 extension ModelExtension on Model {
-  /// Returns the path in the model to [member], or `null` if [member] is not in this `Model`.
+  /// Returns the path in the model to [member], or `null` if [member] is not
+  /// in this [Model].
   ///
-  /// TODO(davidmorgan): this works for any node, but it's not clear yet which types of
-  /// node we want this functionality exposed for.
-  /// TODO(davidmorgan): a list of path segments is probably more useful than `String`.
+  /// Comparison is by identity, not by value, so the exact instance must be in
+  /// this [Model].
+  ///
+  /// TODO(davidmorgan): this works for any node, but it's not clear yet which
+  /// types of node we want this functionality exposed for.
+  /// TODO(davidmorgan): a list of path segments is probably more useful than
+  /// `String`.
   String? pathToMember(Member member) => _pathTo(member.node);
 
-  /// Returns the path in the model to [node], or `null` if [node] is not in this `Model`.
+  /// Returns the path in the model to [node], or `null` if [node] is not in this [Model].
   String? _pathTo(Map<String, Object?> node) {
     if (node == this.node) return '';
     final parent = _getParent(node);
@@ -29,7 +34,7 @@ extension ModelExtension on Model {
     for (final entry in parent.entries) {
       if (entry.value == node) {
         final parentPath = _pathTo(parent);
-        return parentPath == null ? null : '${_pathTo(parent)}/${entry.key}';
+        return parentPath == null ? null : '$parentPath/${entry.key}';
       }
     }
     return null;
@@ -39,19 +44,19 @@ extension ModelExtension on Model {
   Map<String, Object?>? _getParent(Map<String, Object?> node) {
     // If both maps are in the same `JsonBufferBuilder` then the parent is
     // immediately available.
-    if (this is MapInBuffer && node is MapInBuffer) {
-      final thisMapInBuffer = this as MapInBuffer;
-      final thatMapInBuffer = node as MapInBuffer;
-      if (thisMapInBuffer.buffer == thatMapInBuffer.buffer) {
-        return thatMapInBuffer.parent;
+    if (this case MapInBuffer thisMapInBuffer) {
+      if (node case MapInBuffer thatMapInBuffer) {
+        if (thisMapInBuffer.buffer == thatMapInBuffer.buffer) {
+          return thatMapInBuffer.parent;
+        }
       }
     }
     // Otherwise, build a `Map` of references to parents and use that.
-    return _parentsMap[node];
+    return _lazyParentsMap[node];
   }
 
   /// Gets a `Map` from values to parent `Map`s.
-  Map<Map<String, Object?>, Map<String, Object?>> get _parentsMap {
+  Map<Map<String, Object?>, Map<String, Object?>> get _lazyParentsMap {
     var result = _parentsMaps[this];
     if (result == null) {
       result =
@@ -65,12 +70,16 @@ extension ModelExtension on Model {
   static void _buildParentsMap(Map<String, Object?> parent,
       Map<Map<String, Object?>, Map<String, Object?>> result) {
     for (final child in parent.values.whereType<Map<String, Object?>>()) {
-      result[child] = parent;
-      _buildParentsMap(child, result);
+      if (result.containsKey(child)) {
+        throw StateError(
+            'Same node found twice.\n\nChild:\n$child\n\nParent:\n$parent');
+      } else {
+        result[child] = parent;
+        _buildParentsMap(child, result);
+      }
     }
   }
 }
 
 /// Expando storing a `Map` from values to parent `Map`s.
-final Expando<Map<Map<String, Object?>, Map<String, Object?>>> _parentsMaps =
-    Expando();
+final _parentsMaps = Expando<Map<Map<String, Object?>, Map<String, Object?>>>();
--- a/pkgs/dart_model/lib/src/dart_model.dart
+++ b/pkgs/dart_model/lib/src/dart_model.dart
@@ -24,20 +24,36 @@ extension ModelExtension on Model {
   /// types of node we want this functionality exposed for.
   /// TODO(davidmorgan): a list of path segments is probably more useful than
   /// `String`.
-  String? pathToMember(Member member) => _pathTo(member.node);
+  QualifiedName? qualifiedNameOfMember(Member member) =>
+      _qualifiedNameOf(member.node);
 
-  /// Returns the path in the model to [node], or `null` if [node] is not in this [Model].
-  String? _pathTo(Map<String, Object?> node) {
-    if (node == this.node) return '';
-    final parent = _getParent(node);
-    if (parent == null) return null;
-    for (final entry in parent.entries) {
-      if (entry.value == node) {
-        final parentPath = _pathTo(parent);
-        return parentPath == null ? null : '$parentPath/${entry.key}';
-      }
+  /// Returns the [QualifiedName] in the model to [node], or `null` if [node] is not in this [Model].
+  QualifiedName? _qualifiedNameOf(Map<String, Object?> node) {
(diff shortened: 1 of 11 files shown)


pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
...
+    return null;
+  }
+
+  /// Gets the `Map` that contains [node], or `null` if there isn't one.
+  Map<String, Object?>? _getParent(Map<String, Object?> node) {
+    // If both maps are in the same `JsonBufferBuilder` then the parent is
+    // immediately available.
+    if (this is MapInBuffer && node is MapInBuffer) {
+      final thisMapInBuffer = this as MapInBuffer;
+      final thatMapInBuffer = node as MapInBuffer;
```
jakemac53, line 44: I am surprised we don't get type promotion here?

pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
...
+    }
+    return null;
+  }
+
+  /// Gets the `Map` that contains [node], or `null` if there isn't one.
+  Map<String, Object?>? _getParent(Map<String, Object?> node) {
+    // If both maps are in the same `JsonBufferBuilder` then the parent is
+    // immediately available.
+    if (this is MapInBuffer && node is MapInBuffer) {
+      final thisMapInBuffer = this as MapInBuffer;
```
jakemac53, line 43: I would create a local variable (`self`?) or use a pattern match to create a variable for this block, to avoid this cast and rely on promotion instead.

pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
...
+      final thatMapInBuffer = node as MapInBuffer;
+      if (thisMapInBuffer.buffer == thatMapInBuffer.buffer) {
+        return thatMapInBuffer.parent;
+      }
+    }
+    // Otherwise, build a `Map` of references to parents and use that.
+    return _parentsMap[node];
```
jakemac53, line 50: This doesn't look like it builds anything, I would probably just make it a function, or call it `lazyParentsMap` or something like that.

pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
...
+  }
+
+  /// Gets a `Map` from values to parent `Map`s.
+  Map<Map<String, Object?>, Map<String, Object?>> get _parentsMap {
+    var result = _parentsMaps[this];
```
jakemac53, line 55: nit: use `putIfAbsent`

pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
...
+      _buildParentsMap(node, result);
+    }
+    return result;
+  }
+
+  /// Builds a `Map` from values to parent `Map`s.
+  static void _buildParentsMap(Map<String, Object?> parent,
+      Map<Map<String, Object?>, Map<String, Object?>> result) {
+    for (final child in parent.values.whereType<Map<String, Object?>>()) {
+      result[child] = parent;
```
jakemac53, line 68: Is it possible this already exists in some cases (I think it is?). Consider a `putIfAbsent` here which only recursively calls `_buildParentsMap` in the absent case.

pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
...
+      _buildParentsMap(child, result);
+    }
+  }
+}
+
+/// Expando storing a `Map` from values to parent `Map`s.
+final Expando<Map<Map<String, Object?>, Map<String, Object?>>> _parentsMaps =
+    Expando();
```
jakemac53, line 76:
```suggestion
final _parentsMaps = Expando<Map<Map<String, Object?>, Map<String, Object?>>>();
```

pkgs/dart_model/lib/src/json_buffer/closed_map.dart
```diff
@@ -101,16 +106,26 @@ class _ClosedMap
...
   }
+
+  @override
+  bool operator ==(Object other) =>
+      other is _ClosedMap &&
+      other.buffer == buffer &&
+      other._pointer == _pointer;
+
+  @override
+  int get hashCode => buffer.hashCode ^ _pointer.hashCode;
```
jakemac53, line 117:
```suggestion
  int get hashCode => Object.hash(buffer, _pointer.hashCode);
```

pkgs/dart_model/lib/src/json_buffer/iterables.dart
```diff
@@ -39,3 +39,13 @@ class _IteratorFunctionIterable<T> extends Iterable<T> {
   @override
   Iterator<T> get iterator => _function();
 }
+
+/// A `Map` in a `JsonBufferBuilder`.
+abstract interface class MapInBuffer {
```
jakemac53, line 44: Fwiw we could also use Expandos for this, which would also allow us to record parents for other types of maps (and wouldn't need this extra interface and associated type checks).

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
```diff
@@ -347,6 +351,15 @@ class _TypedMap
...
   }
+
+  @override
+  bool operator ==(Object other) =>
+      other is _TypedMap &&
+      other.buffer == buffer &&
+      other._pointer == _pointer;
+
+  @override
+  int get hashCode => buffer.hashCode ^ _pointer.hashCode;
```
jakemac53, line 362:
```suggestion
  int get hashCode => Object.hash(buffer, _pointer);
```

pkgs/dart_model/test/model_test.dart
```diff
@@ -107,6 +107,34 @@ void main() {
           expected['uris']!['package:dart_model/dart_model.dart']!['scopes']![
               'JsonData']!['members']);
     });
+
+    test('can give the path to Members in buffer backed maps', () {
+      final member = model.uris['package:dart_model/dart_model.dart']!
+          .scopes['JsonData']!.members['_root']!;
+      expect(model.pathToMember(member),
+          '/uris/package:dart_model/dart_model.dart/scopes/JsonData/members/_root');
```
jakemac53, line 115: This doesn't look like a very useful path, how are you supposed to parse out the library URI? What is the intended use case for these paths?

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: Neither promote works because neither `Model` nor `Map<String, Object?>` have any common supertype with `MapInBuffer`.

But `case` works, my first `if case` :) done.

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: Done.

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: Aha, the one time my laziness pays off ;) ... `Expando` apparently doesn't have `putIfAbsent`.

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: It shouldn't: every node should have a unique path, with no cycles and no reuse. Otherwise using paths like this doesn't really work.

Added a check to throw if there's a duplicate, and tests to check it throws for a cycle or a duplicate.

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: Done.

pkgs/dart_model/lib/src/json_buffer/closed_map.dart
davidmorgan: Done, and in the two other maps.

pkgs/dart_model/lib/src/json_buffer/iterables.dart
davidmorgan: Good point, thanks.

Another way to tweak would be to lean on the type more, the parent of a map in a buffer is definitely also a map in a buffer. So there should only be one type check to get the whole path. (And as noted elsewhere, we should store the key used as well as the parent, so it really is fast).

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
davidmorgan: Done.

pkgs/dart_model/test/model_test.dart
davidmorgan: Mentioned elsewhere :) indeed I expect it will change to path segments or maybe something even more specific like a lookup of exactly library URI.

It is intended for augmentation builders which we don't have quite yet :)

pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -11,3 +12,65 @@ extension QualifiedNameExtension on QualifiedName {
 
   bool equals(QualifiedName other) => other.uri == uri && other.name == name;
 }
+
+extension ModelExtension on Model {
+  /// Returns the path in the model to [member], or `null` if [member] is not in this `Model`.
+  ///
+  /// TODO(davidmorgan): this works for any node, but it's not clear yet which types of
+  /// node we want this functionality exposed for.
+  /// TODO(davidmorgan): a list of path segments is probably more useful than `String`.
```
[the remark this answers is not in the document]
jakemac53: I think specifically, we just want to be able to get a qualified name for any given member? Can this just return that instead?

pkgs/dart_model/lib/src/dart_model.dart
jakemac53: The case I was thinking of is two different nodes which are siblings both ask for parents, they will both end up re-computing the parents of their parents?

jakemac53 (approved):

pkgs/dart_model/lib/src/dart_model.dart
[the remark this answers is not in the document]
davidmorgan: Done.

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: I missed replying to this yesterday, sorry about that.

Actually that's not a problem: the parents map is always computed for the whole `Model` if it's needed; then it won't be computed again. So no re-computation :)

PR #89, opened 2024-10-08
@JsonCodable() supports super toJson and fromJson.

davidmorgan:
Add checking of the super type for `fromJson` and `toJson`, which requires adding query support for constructors and methods.

I just added them flat in `Member`, there is still the TODO to add more types there, this gives something useful for new types to do, i.e. differentiate constructors, methods and fields :)

This also needs to look up the supertype, @simolus3 I wasn't quite sure where that should go, added to `StaticTypeSystem` [here](https://github.com/dart-lang/macros/pull/89/files#diff-28da2a69e681ea5746235dd3017c9c7ca0c0dc6c1fb535bcdd76d3310f54fdf6), or did you have something else in mind? :)

### DIFF
--- a/pkgs/_test_macros/lib/json_codable.dart
+++ b/pkgs/_test_macros/lib/json_codable.dart
@@ -44,16 +44,30 @@ class JsonCodableImplementation implements Macro {
   }
 
   Future<AugmentResponse> phase3(Host host, AugmentRequest request) async {
-    final result = <Augmentation>[];
-
     final target = request.target;
     final model = await host.query(Query(target: target));
     final clazz = model.uris[target.uri]!.scopes[target.name]!;
 
+    // TODO(davidmorgan): put `extends` information directly in `Interface`.
     MacroScope.current.addModel(model);
+
+    return AugmentResponse(augmentations: [
+      await _generateFromJson(host, model, target, clazz),
+      await _generateToJson(host, model, target, clazz)
+    ]);
+  }
+
+  Future<Augmentation> _generateFromJson(
+    Host host,
+    Model model,
+    QualifiedName target,
+    Interface clazz,
+  ) async {
+    // TODO(davidmorgan): put `extends` information directly in `Interface`.
     final superclassName = MacroScope.current.typeSystem.supertypeOf(target);
 
     var superclassHasFromJson = false;
+    // TODO(davidmorgan): add recommended way to check for core types.
     if (superclassName.asString != 'dart:core#Object') {
       // TODO(davidmorgan): first query could already fetch the super class.
       final supermodel = await host.query(Query(target: superclassName));
@@ -86,10 +100,20 @@ class JsonCodableImplementation implements Macro {
 
     // TODO(davidmorgan): helper for augmenting initializers.
     // See: https://github.com/dart-lang/sdk/blob/main/pkg/_macros/lib/src/executor/builder_impls.dart#L500
-    result.add(Augmentation(code: '''
+    return Augmentation(code: '''
 augment ${target.name}.fromJson($_jsonMapType json) :
 ${initializers.join(',\n')};
-'''));
+''');
+  }
+
+  Future<Augmentation> _generateToJson(
+    Host host,
+    Model model,
+    QualifiedName target,
+    Interface clazz,
+  ) async {
+    // TODO(davidmorgan): put `extends` information directly in `Interface`.
+    final superclassName = MacroScope.current.typeSystem.supertypeOf(target);
 
     var superclassHasToJson = false;
     if (superclassName.asString != 'dart:core#Object') {
@@ -135,6 +159,8 @@ ${serializers.join('')}
 
     return AugmentResponse(augmentations: result);
   }
+
+  String _generateFromJson() {}
 
   /// Returns whether [constructor] is a constructor
   /// `fromJson(Map<String, Object?>)`.
--- a/pkgs/_test_macros/lib/json_codable.dart
+++ b/pkgs/_test_macros/lib/json_codable.dart
@@ -49,11 +49,11 @@ class JsonCodableImplementation implements Macro {
     final clazz = model.uris[target.uri]!.scopes[target.name]!;
 
     // TODO(davidmorgan): put `extends` information directly in `Interface`.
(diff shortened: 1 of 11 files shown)


pkgs/dart_model/lib/src/type_system.dart
```diff
@@ -61,6 +61,14 @@ final class StaticTypeSystem {
     return Scope.none.run(() => _isSubtype(a, b));
   }
 
+  /// Returns the super type of the type referred to by [name], which must be
+  /// part of the model.
+  ///
+  /// TODO(davidmorgan): is this the right place to provide access to supertypes?
+  QualifiedName supertypeOf(QualifiedName name) {
+    return _constructSuperTypes(_lookupNamed(name.asString), []).first.name;
```
simolus3, line 69: For types, I think this is the right place to put it. A `QualifiedName` alone is not really a type though, interface types would also need the instantiation. And this method answers class hierarchy queries, which are not exactly the same as type hierarchy queries:

1. There is no designated "primary" supertype in the type hierarchy since it's not relevant when checking for subtype relationships. Using `.first` here essentially relies on an implementation detail.
2. This generates an intermediate fully-resolved parent type when we only care about the name (so at the very least has to run in the none scope I think?)

I absolutely think that they're should be methods to resolve super types of interface _types_ in `TypeSystem`. I'm not sure if the `QualifiedName` is close enough to an interface type to also belong here though.
Especially if we care about the type of subclass relationship, maybe it's cleaner to add `extends`, `implements` and `mixesIn` properties on `Interface` as well?

In my mind, the kinds of question a type system should answer are "is this class, through whatever how many intermediates, a subtype of `JsonConverter`?". IMHO, "does this class extend from a class defining a `fromJson` constructor" is a different kind of question that is more related to the AST instead of a type model. But that might just be my opinion, so feel free to disagree or ignore this if it doesn't make sense :)

schemas/dart_model.schema.json
```diff
@@ -117,6 +117,31 @@
         "returnType": {
           "$comment": "The return type of this member, if it has one.",
           "$ref": "#/$defs/StaticTypeDesc"
+        },
+        "requiredPositionalParameters": {
```
jakemac53, line 121: I wouldn't separate the required from the optional positional parameters - I would look at the existing introspection APIs for declarations when adding them here, for example https://github.com/dart-lang/sdk/blob/main/pkg/_macros/lib/src/api/introspection.dart#L269. These have been fairly well fleshed out at this point and should be directly copied imo.

schemas/dart_model.schema.json
```diff
@@ -117,6 +117,31 @@
...
+        "optionalPositionalParameters": {
+          "type": "array",
+          "description": "The optional positional parameters of this member, if it has them.",
+          "items": {
+            "$ref": "#/$defs/StaticTypeDesc"
+          }
+        },
+        "namedParameters": {
+          "type": "array",
+          "description": "The named positional parameters of this member, if it has them.",
```
jakemac53, line 137:
```suggestion
          "description": "The named parameters of this member, if it has them.",
```

schemas/dart_model.schema.json
```diff
@@ -117,6 +117,31 @@
...
+          "items": {
+            "$ref": "#/$defs/NamedFunctionTypeParameter"
+          }
+        },
+        "parameters": {
```
jakemac53, line 142: I am confused by this one, what does it represent?

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -49,7 +49,28 @@ class JsonCodableImplementation implements Macro {
     final target = request.target;
     final model = await host.query(Query(target: target));
     final clazz = model.uris[target.uri]!.scopes[target.name]!;
-    // TODO(davidmorgan): check for super `fromJson`.
+
+    MacroScope.current.addModel(model);
```
jakemac53, line 53: Fwiw, just reading this code I don't have any idea what it does, I have to go read the API docs which isn't great.

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -49,7 +49,28 @@ class JsonCodableImplementation implements Macro {
...
+    final superclassName = MacroScope.current.typeSystem.supertypeOf(target);
+
+    var superclassHasFromJson = false;
+    if (superclassName.asString != 'dart:core#Object') {
```
jakemac53, line 57: Just a comment on this API as well, the string comparison here seems less than ideal. It also is going to have issues if for instance an identifier is being re-exported from some other library. We probably need a better way to do this. This type of check is only going to work for core libraries really.

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -49,7 +49,28 @@ class JsonCodableImplementation implements Macro {
...
+      // TODO(davidmorgan): first query could already fetch the super class.
+      final supermodel = await host.query(Query(target: superclassName));
```
jakemac53, line 59: We should also add a TODO for a more efficient type of query here

pkgs/_test_macros/lib/json_codable.dart
```diff
@@ -49,7 +49,28 @@ class JsonCodableImplementation implements Macro {
     final target = request.target;
     final model = await host.query(Query(target: target));
     final clazz = model.uris[target.uri]!.scopes[target.name]!;
-    // TODO(davidmorgan): check for super `fromJson`.
+
+    MacroScope.current.addModel(model);
+    final superclassName = MacroScope.current.typeSystem.supertypeOf(target);
+
+    var superclassHasFromJson = false;
```
jakemac53, line 56: nit: I think it would help readability to extract the toJson/fromJson parts into separate functions

pkgs/dart_model/lib/src/type_system.dart
davidmorgan: All good points, thanks! I agree re: "what does this extend" being more like an AST question than a type model, I think moving this functionality to `Interface` is the right direction.

There's already a lot going on in this PR so added a TODO pointing to this comment + an item to the short term planning issue.

jakemac53 (approved):

Issue #92, opened 2024-10-09
Support for nested collections in schema generator


jakemac53:
I wanted to add a property of type `Map<String, List<Augmentation>>` but this doesn't work today. It is pretty trivial to get the generator in a state where it will at least not crash, but the getters it generates in the extension types just call `.cast()` on the map which won't work for the values, which also need to be casted.

We might need our own map implementation for this to do lazy deep casting, I don't see any appropriate methods on Map to support it (`.map` could work but it wouldn't be lazy).

davidmorgan:
After #76 one `Augmentation` already contains a `List<Code>`, maybe that gives you enough flexibility to not need an outer `List`?

Agreed re: casting. Seems okay to add if it looks useful, we can always revisit.

Thanks.

PR #95, opened 2024-10-10
Add support for deep collection casting in the generator

jakemac53:
Closes https://github.com/dart-lang/macros/issues/92

- Adds a `deepCast` extension on maps which takes a function for deep casting values.
- Add support in the generator for using it.
- As a part of this, it also explicitly passes type arguments to `cast` and `deepCast` always, instead of relying on inference.

There is no testing of the generator code itself, since it is only tested by essentially validating actual output, but I manually tested it, and it will be used in a follow-up.

### DIFF
--- a/tool/dart_model_generator/lib/generate_dart_model.dart
+++ b/tool/dart_model_generator/lib/generate_dart_model.dart
@@ -66,6 +66,7 @@ class GenerationContext {
     }.nonNulls;
     return ([
       "import 'package:dart_model/src/json_buffer/json_buffer_builder.dart';",
+      "import 'package:dart_model/src/deep_cast_map.dart';",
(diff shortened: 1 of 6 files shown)


tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -466,6 +481,7 @@ class ClassTypeDefinition implements Definition {
       result.writeln('  $name() : ');
     } else {
       result.writeln('  $name({');
+      // TODO: Why are we excluding Map properties?
```
davidmorgan, line 484: It's to stop people passing them in as SDK maps that we then have to copy :) I'll send you a follow-up explaining.

davidmorgan (approved):
Looks great, thanks!

Issue #96, opened 2024-10-10
QualifiedName should be able to reference top level members of libraries and members of interfaces


jakemac53:
It seems to me today that it can only reference Interfaces within a library, so we will have to figure out how it needs to change, and what the URI syntax of it should be (but, I think we will find this is hard to represent as a URI, fwiw).

We also need to be able to target whole libraries, so if we use QualifiedName to represent the "target" in macro APIs (as it does now), then they also need to be able to represent the library itself. But, if we move to specific URIs for each phase/target combo, we could just use a Uri instead as the target when running on a library.

davidmorgan:
`Model` doesn't actually have a place for top level members yet, we need to decide whether block declarations and members are part of the same union ... in which case top level members can just be referenced by name in the existing map ... or whether there is a special "top level" block declaration with members, which could either get its own field in the model or be in the existing `Map` and named the empty String `''`.

If we choose to describe top level members as part of a block declaration called `''` in every library then referring to it with a `QualifiedName(name: '')` sort of makes sense.

Taking a step back, though: it might make sense to introduce a new union type for "macro target"? Then we can certainly target all the different types of thing :)

jakemac53:
I am assuming that `QualifiedName` is the full replacement for `Identifier`, which would mean it also is what you use in code to emit references to things. That would also have the same requirements here of being able to reference any declaration at any scope.

So, I think we should figure out how to allow it to do that. Maybe adding a `Scope` instead of a `Uri`? I am not sure we should probably chat about it.

davidmorgan:
I do not have a clear idea the best way to go here :)

We have the capability to use any node in the `Model` for referring to things, so for example we can use `Member` and turn that into `QualifiedName` + member name.

Maybe what we are actually talking about is a path into `Model`? There is certainly a lot we can do there and have it be fully general and extensible, which is interesting :) ... for example I think we can generate metadata so that a path into the model acts as a union over all the types we have; you can ask it what type it is, and retrieve the actual value with that static type.

Given we are still free to refactor, it probably works okay all ways around: we could expand `QualifiedName` and later split it into more/different types; or start with new types and later combine them if they turn out to be unnecessary. And/or we can explore the possibilities around general paths.

davidmorgan:
A couple of notes from our chat:

- Path isn't sufficient, since you can only refer to things that are already in `Model`; you might want to reference something you haven't generated yet.
[...]

PR #101, opened 2024-10-11
add scope and isStatic to QualifiedName, as well as parsing and a URI format

jakemac53:
Closes https://github.com/dart-lang/macros/issues/96, I think for libraries we will want just a Uri essentially, since they aren't first class objects in Dart, you can't refer directly to them.

### DIFF
--- a/tool/dart_model_generator/lib/definitions.dart
+++ b/tool/dart_model_generator/lib/definitions.dart
@@ -237,21 +237,59 @@ static Protocol handshakeProtocol = Protocol(
                   type: 'bool', description: 'Whether the entity is static.'),
             ]),
         Definition.clazz('QualifiedName',
-            description: 'A URI combined with a name.',
+            description: 'A URI combined with a name and scope referring to a '
+                'declaration. The name and scope are looked up in the export '
+                'scope of the URI.',
             createInBuffer: true,
             properties: [
               Property('uri',
                   type: 'String',
                   description: 'The URI of the file containing the name.'),
-              Property('name', type: 'String', description: 'The name.'),
+              Property('scope',
+                  type: 'String',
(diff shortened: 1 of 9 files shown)


tool/dart_model_generator/lib/definitions.dart
```diff
@@ -237,21 +237,59 @@ static Protocol handshakeProtocol = Protocol(
...
                   description: 'The URI of the file containing the name.'),
-              Property('name', type: 'String', description: 'The name.'),
+              Property('scope',
+                  type: 'String',
+                  description: 'The optional scope to look up the name in.',
+                  nullable: true),
+              Property('name',
+                  type: 'String',
+                  description: 'The name of the declaration to look up.'),
+              Property('isStatic',
```
davidmorgan, line 255: I guess this will probably end up as an enum, since we might want to distinguish more types of thing (e.g. getter vs setter vs field), and more might arrive later. Maybe a TODO since it's easy to change for now?

davidmorgan (approved):
Looks good, thanks :)

tool/dart_model_generator/lib/definitions.dart
jakemac53: I don't think it ever matters whether it is a getter/setter/field for the purposes of this object, it is just telling you how to look up a declaration unambiguously, which does require knowing if it should be looked up in the static or instance member scope.

davidmorgan: Doesn't matter for right now, but I was thinking of a case where you have e.g. a subclass with a setter and superclass with a getter; you need to say setter or getter to disambiguate between the two. It's a bit of an awkward corner :) anyway, we can figure that out later if needed. Thanks.

Issue #100, opened 2024-10-11
AugmentRequest should contain basic information about the declaration


jakemac53:
Today all you get is a QualifiedName for the target, but we now have to do a query before even calling into the macro functions, since we don't even know what kind of declaration it is any more.

We should instead provide the declaration, but no members of the declaration, to the macro up front.

Because of the query model, the macro will end up getting some duplicate information later on if they do any additional queries for the members, but I think that is probably OK.

Issue #103, opened 2024-10-15
Consider mimicking type hierarchies in our extension types.


jakemac53:
Specifically the idea here would be to add an `String? implements` option to the generator for class definitions. If present, the extension type would implement the given type.

All properties from that implemented type would be copied into the extension types constructors and schema. The members would not be regenerated though, since they would come with the implements clause.

Specifically the first thing we would use this for is to create a `Declaration` type, which has the shared properties of all declarations. And then the API for getting a qualified name for a node in a model could be `QualifiedName? qualifiedNameOf(Declaration declaration)`, as opposed to needing specialized methods for each declaration type, or making you explicitly call `.node` and accepting a Map type (I have a separate change to do the latter out pending).

I am not exactly sure how to represent these in the schema in the best way (union types, and copy the "inherited" members into each schema type?).

PR #104, opened 2024-10-15
Pass target model in AugmentRequest

jakemac53:
Closes https://github.com/dart-lang/macros/issues/100

- Adds `model` field to AugmentRequest, with implementations to pass it (although, these use a host side query).
  - Expose the `hostService` from the `MacroHost` (used to do the query)
- Adds a `Model` parameter to all macro methods (typed as `Object` for right now for most of them, as we don't have better types yet).
- Make all fields on AugmentRequest required (this simplified implementations, but lmk if you want me to revert this).
- Rename the `Model.qualifiedNameOfMember` extension to `qualifiedNameOf` generalizing and fixing it:
  -  Now it actually does what it says 🤣 (it used to return just the qualified name of the type the member is declared in I think?), there was in any case no way to represent members with qualified names, see https://github.com/dart-lang/macros/pull/101 which adds that.
  - It now throws for anything other than types, since they can't be referenced
  - It just accepts a `Map<String, Object?>` now which is the only thing we can generalize on for the moment, but see https://github.com/dart-lang/macros/issues/103 for one possible workaround.
- Fixes up tests and removes expected request/response from the client for the model

### DIFF
--- a/pkgs/_analyzer_macros/lib/macro_implementation.dart
+++ b/pkgs/_analyzer_macros/lib/macro_implementation.dart
@@ -84,12 +84,13 @@ class AnalyzerRunningMacro implements injected.RunningMacro {
   /// Queries for [target] and returns the [Model] representing the result.
   ///
   /// TODO: Make this a more limited query which doesn't fetch members.
-  Future<Model> _queryTarget(QualifiedName target) async =>
-      (await _impl._host.hostService.handle(MacroRequest.queryRequest(
-              QueryRequest(query: Query(target: target)),
-              id: nextRequestId)))
+  Future<Model> _queryTarget(QualifiedName target) =>
+      Scope.query.run(() async => (await _impl._host.hostService.handle(
+              MacroRequest.queryRequest(
+                  QueryRequest(query: Query(target: target)),
+                  id: nextRequestId)))
           .asQueryResponse
-          .model;
+          .model);
 
   @override
   Future<AnalyzerMacroExecutionResult> executeDeclarationsPhase(
--- a/pkgs/_analyzer_macros/lib/macro_implementation.dart
+++ b/pkgs/_analyzer_macros/lib/macro_implementation.dart
@@ -81,6 +81,16 @@ class AnalyzerRunningMacro implements injected.RunningMacro {
     return AnalyzerRunningMacro._(impl, name, implementation);
   }
 
+  /// Queries for [target] and returns the [Model] representing the result.
+  ///
+  /// TODO: Make this a more limited query which doesn't fetch members.
+  Future<Model> _queryTarget(QualifiedName target) async =>
+      (await _impl._host.hostService.handle(MacroRequest.queryRequest(
+              QueryRequest(query: Query(target: target)),
+              id: nextRequestId)))
+          .asQueryResponse
+          .model;
+
   @override
   Future<AnalyzerMacroExecutionResult> executeDeclarationsPhase(
       macros_api_v1.MacroTarget target,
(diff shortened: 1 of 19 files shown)


jakemac53:
I am not able to figure out how to fix the remaining test failures related to the scopes in which these models are created :/.

I tried sprinkling a bunch of `Scope.query.run` stuff around various parts but to no avail.

davidmorgan:
Thanks! I think there are three things worth discussing here, maybe we can discuss today?

## Minimizing data + round trips.

I think we need an API design here that is perfect for performance, or very close. 

That means no unnecessary data returned and no unnecessary round trips.

I think the way to do that is for the macro to specify its initial query in the `MacroDescription`. We still need to flesh out queries, but I guess it can be something like: the `MacroDescription` has the initial query with missing `target`, and the host fills in the target.

This could be implemented as a next step after this PR, but there is one aspect that is maybe worth thinking about now, which is that we then need API for specifying the initial query, for example a getter that goes with each macro method, for the bootstrap code to call.

## Evolving classes vs parameters.

Classes are easier to evolve than lists of parameters, because you can add fields to a class and old code that receives the class does not have to change. So I wonder if we should put as much as we can in one "macro scope" object ... actually, we have scopes, see next item ;)

That would leave the target as the only parameter for these methods, which makes sense because it's the thing that can be different per method. So for example `buildDefinitionsForClass(Interface target)`. We would never change `Interface` to a different type, we would add a new method that accepts the different type.

## Use of `Scope` including implicit use.

I realize now that we use the word "scope" twice :( ... let's change that ... here I mean the scope that is a dependency injection scope, not the Dart language concept.

We have a concept of macro scope with a `Model` in it, so we can just put the model there and not pass it explicitly.

`Host` seems to fit in with that, i.e. a macro always has a `Host` available to it, we can add it to the scope.

For implicit use, I mean for example that `qualifiedName = model.qualifiedNameOf(target.node)` can be replaced with getters like `Interface.qualifiedName` that get the model from the scope in order to look up the qualified name.

jakemac53:
> I think the way to do that is for the macro to specify its initial query in the `MacroDescription`. We still need to flesh out queries, but I guess it can be something like: the `MacroDescription` has the initial query with missing `target`, and the host fills in the target.

We might want them to be able to only request certain aspects of the `target` as well, as opposed to getting an implicit query with all the header info (this way you could say, add a mixin and not re-run the macro if it doesn't care about mixins).

I am a bit skeptical that they will be able to express the query at all without already having access to the target. So, we might need to invent a special syntax for expressing queries as templates with variables, in this case the variable would be the target. But, macros can also target different kinds of targets so I am still a bit skeptical this will work out well. We will have to work through examples of that.

I do think it makes sense for the API to provide the target object directly still and not just a generic Model for the initial query response. 

> This could be implemented as a next step after this PR, but there is one aspect that is maybe worth thinking about now, which is that we then need API for specifying the initial query, for example a getter that goes with each macro method, for the bootstrap code to call.

I do think it should be a next step, this PR can be adapted to fit whatever we come up with.

> For implicit use, I mean for example that `qualifiedName = model.qualifiedNameOf(target.node)` can be replaced with getters like `Interface.qualifiedName` that get the model from the scope in order to look up the qualified name.

How would we deal with models from query responses? Would the implicit model from the scope always get merged with each query response (maybe using https://github.com/dart-lang/macros/pull/105?)

davidmorgan (approved):
LG to land with various followups as discussed :) thanks.

jakemac53:
Specific follow up summary, will convert these 

- land this pr as is
- let you define your initial query
- Make the API only take a builder object, with access to the target model via a getter (@davidmorgan we agreed this should be the model (ie: Interface) and not the qualified name right?)
- Add some convenience methods on qualified name to look it up in the model because this will be generally useful.

davidmorgan:
Yes, I think the builder can know which type is useful and return that specific type instead of `QualifiedName`.

Issue #102, opened 2024-10-14
Add ability to produce a "merged" model instance


jakemac53:
You might have multiple queries and want to merge the results of them together as opposed to duplicating them. We should be able to do this fairly cheaply by creating a `MergedMap` class which just takes two (or more) maps. As you navigate down it, any path which exists in both models implicitly creates a new "MergedMap" class with the results from each sub-map. It won't be free but it should be fairly ergonomic.

Thoughts? cc @davidmorgan

PR #105, opened 2024-10-17
Add LazyMergedMap for merging Model objects

jakemac53:
Possibly resolution to https://github.com/dart-lang/macros/issues/102, not high priority to land right away just putting something out there.

Has a few failing tests right now which I can look at if we want to move forward.

We might want to benchmark this also if we decide to land it.

### DIFF
--- a/pkgs/dart_model/lib/src/lazy_merged_map.dart
+++ b/pkgs/dart_model/lib/src/lazy_merged_map.dart
@@ -30,6 +30,7 @@ class LazyMergedMapView extends MapBase<String, Object?> {
 
   @override
   Object? operator [](Object? key) {
+    // TODO: Can we do better? These lookups can each be linear for buffer maps.
     var leftValue = left[key];
     var rightValue = right[key];
     if (leftValue != null) {
--- a/pkgs/dart_model/lib/src/lazy_merged_map.dart
+++ b/pkgs/dart_model/lib/src/lazy_merged_map.dart
@@ -72,6 +72,18 @@ class LazyMergedMapView extends MapBase<String, Object?> {
       throw UnsupportedError('Merged maps are read only');
 }
 
+extension AllMaps on Map<String, Object?> {
+  /// All the maps merged into this map, recursively expanded.
+  Iterable<Map<String, Object?>> get expand sync* {
+    if (this case final LazyMergedMapView self) {
+      yield* self.left.expand;
+      yield* self.right.expand;
+    } else {
+      yield this;
+    }
+  }
+}
+
 extension MergeModels on Model {
   /// Creates a lazy merged view of `this` with [other].
   Model mergeWith(Model other) =>
--- a/pkgs/dart_model/lib/src/lazy_merged_map.dart
+++ b/pkgs/dart_model/lib/src/lazy_merged_map.dart
@@ -1,0 +1,78 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
+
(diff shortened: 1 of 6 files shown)


pkgs/dart_model/test/lazy_merged_map_test.dart
```diff
@@ -0,0 +1,66 @@
...
+    final c = a.mergeWith(b);
+    final properties = c.uris['package:a/a.dart']!.scopes['A']!.properties;
+    expect(properties, isA<LazyMergedMapView>());
+    expect(properties.isClass, true);
+    expect(properties.isAbstract, true);
+    // Not set
+    expect(() => properties.isConstructor, throwsA(isA<TypeError>()));
+  });
+
+  test('Errors if identical keys are not equal', () async {
```
davidmorgan, line 52: test name: "identical"/"equal" made me think this is about `identical` vs `operator==`, which is misleading? So maybe something like: "Errors if maps have same key with different values"?

pkgs/dart_model/lib/src/lazy_merged_map.dart
```diff
@@ -0,0 +1,78 @@
...
+/// The [keys] getter will de-duplicate the keys.
+class LazyMergedMapView extends MapBase<String, Object?> {
+  final Map<String, Object?> left;
+  final Map<String, Object?> right;
+
+  LazyMergedMapView(this.left, this.right);
+
+  @override
+  Object? operator [](Object? key) {
+    var leftValue = left[key];
```
davidmorgan, line 33: This can be pretty slow if we hit a case where both maps are buffer-backed and have large matching maps, since lookup is linear.

That's fine for now, there are lots of possibilities to optimize, we should do it with realistic use cases. Maybe a TODO for that?

davidmorgan (approved):
Thanks!

Yes, I think we almost certainly want this, the only way I could see avoiding it is if we require that macros only make one query.

There is a TODO for merging the model into the current scope here

https://github.com/dart-lang/macros/blob/a4822883ce98292173923b07288d1b182894781b/pkgs/dart_model/lib/src/scopes.dart#L162

which could directly use this, not sure if that's useful to do in this PR or not.

jakemac53:
> There is a TODO for merging the model into the current scope here

Oh nice, I will add that

pkgs/dart_model/lib/src/lazy_merged_map.dart
jakemac53: SGTM, I am not sure what we could do here exactly though, I think we always have to look up in each map.

davidmorgan: I was thinking more of optimized `values` and `entries` methods than optimized lookup. So it's still fast to just walk the whole `Map`.

Making lookup fast might already be a problem before merging :)

Anyway, something for later. Thanks.

pkgs/dart_model/lib/src/scopes.dart
```diff
@@ -157,10 +157,12 @@ class MacroScope {
 
   /// Merges a new partial [model] into the scope's accumulated model spanning
   /// multiple queries.
-  // TODO: Actually accumulate instead of replacing the model
   void addModel(Model model) {
-    _accumulatedModel = model;
-    _typeSystem = null;
+    if (_accumulatedModel case var accumulated?) {
+      accumulated.mergeWith(model);
```
davidmorgan, line 162: I think that should be `_accumulatedModel = ...` :)

I guess we don't have any test coverage for this :) meaning we never use the type system after a second query.

jakemac53: hah!

jakemac53: https://github.com/dart-lang/macros/pull/110

PR #106, opened 2024-10-17
Make AugmentResponse more like MacroExecutionResult

jakemac53:
In preparation for adding builder types, and making `AugmentResponse` not visible in the macro APIs any more.

Will do that after https://github.com/dart-lang/macros/pull/104 is also merged, as we really need both.

### DIFF
--- a/tool/dart_model_generator/lib/generate_dart_model.dart
+++ b/tool/dart_model_generator/lib/generate_dart_model.dart
@@ -353,7 +353,7 @@ class TypeReference {
       return '($rawCast)$q.deepCast<String, ${elementType!.dartType}>('
           '(v) => ${elementType!.castExpression('v')})';
     } else if (isList) {
-      if (elementType!.elementType == null) return '($rawCast).cast()';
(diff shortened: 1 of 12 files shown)


tool/dart_model_generator/lib/generate_dart_model.dart
```diff
@@ -493,6 +493,7 @@ class ClassTypeDefinition implements Definition {
       result.writeln('  $name() : ');
     } else {
       result.writeln('  $name({');
+      // TODO: Why are we excluding Map properties?
```
davidmorgan, line 496: I think this is a merge mistake? I added a comment for it up above, anyway :)

davidmorgan (approved):
Thanks!

Please see my comment on #104 for some general thoughts also relevant here, but this can be merged as is.

One possible way to proceed afterwards would be to have the methods return `void`, instead of returning a value, augmentations are accumulated in the `Scope`. (Dependency injection scope, not a Dart scope). This avoids mentioning `AugmentationResult` and seems like it should work well with builders.

Issue #108, opened 2024-10-18
Evaluate and investigate performance


jakemac53:
There are many aspects to this, I figure we can collect info in this issue so as not to duplicate too much work.

Some specific tasks are:

- [ ] Establish a benchmarking strategy/suite, may need help from implementation teams but TBD
- [ ] Improve invalidation based on query results (will probably need some work by @scheglov and @johnniwinther )
- [ ] Improve granularity of queries to get more granular invalidation
- [ ] Make improvements once problem areas are identified

jakemac53:
There should be some very low hanging fruit here - I am seeing almost 100ms to run a phase which is doing nothing at all (types phase, for json_codable).

jakemac53:
https://github.com/dart-lang/macros/pull/104 should also improves things here.

scheglov:
We previously discussed invalidation in https://github.com/dart-lang/sdk/issues/55784, and it seems to me that timings provided there will still be the same, even with new queries. The analyzer invalidates and links whole library cycles, and has to parse all files of the library cycle. We cannot currently do it with higher granularity. We still want by UX reasons to have single macro generated file per library. We can avoid running individual macros, or individual phases, but this will save us only so much.

scheglov:
Yes, I agree that queries might simplify the invalidation checks. Could you walk me through a scenario?

Lets say we have the library `L` with two macro applications `M1` and `M2`. Where `M1` on the types phase introspects class `C1` and generates class `C2`, and configures it so that `M2` runs on `C1` during the declarations phase. We can remember the query `Q1` used by `M1`, and so during the next linking of `L`  instead of running `M1` we compute the answer on `Q1`, see that it is the same as before, and know that we can reuse the previously generated `R1`.

However we cannot use previously remembered `Q2` query until we apply `R1` into `L`, because `Q2` might (and did) introspect the results of the previously run macros, `C2` of `M1`. So, we have to parse results of `M1` and integrate partial augmentations into `L`, and only then move to `M2`.

It looks to me that we cannot just run all queries from all macros and decide whether we can reuse the merged result of running macros. We still have to do this macro by macro.

jakemac53:
cc @davidmorgan

Good points, I had forgotten this extra complication of the query results being affected also by macros. So, you need to somehow be able to return the query results as of a specific point in time, which is not only per phase but also even can be affected by individual macros in the phase.

[...]

PR #109, opened 2024-10-18
set tcpNoDelay option on all sockets for 100x speedup

jakemac53:
Related to https://github.com/dart-lang/macros/issues/108, makes round trips about 100x faster.

Total speedup is about 20x, which is still slow (200ms for watch mode edit/refresh loop), but much better.


davidmorgan (approved):
Thanks :)

PR #110, opened 2024-10-22
fix addModel and add test, expose current model

jakemac53:
Follow up to https://github.com/dart-lang/macros/pull/105#discussion_r1810154444


pkgs/dart_model/test/scopes_test.dart
```diff
@@ -7,7 +7,7 @@ import 'package:dart_model/src/json_buffer/json_buffer_builder.dart';
 import 'package:test/test.dart';
 
 void main() {
-  group(Scope, () {
+  group('Scope', () {
```
jakemac53, line 10: Note that I am converting these all to strings because anything else breaks the VsCode integration

PR #113, opened 2024-10-23
Extend typed_map to allow 9-16 fields.

davidmorgan:
Since we might have some bigger types soon :)

### DIFF
--- a/pkgs/dart_model/lib/src/json_buffer/typed_map.dart
+++ b/pkgs/dart_model/lib/src/json_buffer/typed_map.dart
@@ -71,15 +71,24 @@ class TypedMapSchema {
   /// [Type#_sizeInBytes] for all present values.
   ///
   /// Additionally returns a [bool]: whether the map is filled.
-  (int, bool) _valueSizeOf(
-      [Object? v0,
-      Object? v1,
-      Object? v2,
-      Object? v3,
-      Object? v4,
-      Object? v5,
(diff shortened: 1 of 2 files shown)


pkgs/dart_model/lib/src/json_buffer/typed_map.dart
```diff
@@ -187,6 +236,13 @@ extension TypedMaps on JsonBufferBuilder {
       void addBit(bool bit) {
         if (bit) byte += bitmask;
         bitmask <<= 1;
+
+        if (bitmask == 0x100) {
+          _setByte(valuePointer, byte);
```
jakemac53, line 241: byte can be two bytes now right? does _setByte need to be renamed, does it actually work for 2 bytes?

jakemac53 (approved):

pkgs/dart_model/lib/src/json_buffer/typed_map.dart
davidmorgan: Added a comment clarifying

```
// On reaching bit 9 of `byte`, write the byte and reset `byte` and
// `bitmask`.
```

jakemac53: Oh right, I was thinking `0x100` was 16 bytes. We really need binary integer literals lol.

PR #120, opened 2024-10-25
Add an `analysis_server` binary with `dart_model` implementation injected

davidmorgan:
There are definitely some correctness issues, in particular it looks like the cached results from with no macro impl injected prevent macros running, so I had to add some change to trigger macros.

But it does work interactively on e.g. `json_codable_test.dart`.

### DIFF
--- a/pkgs/_analyzer_macros/bin/server.dart
+++ b/pkgs/_analyzer_macros/bin/server.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:io';
-
 import 'package:_analyzer_macros/macro_implementation.dart';
 import 'package:analysis_server/starter.dart';
 import 'package:analyzer/src/summary2/macro_injected_impl.dart' as injected;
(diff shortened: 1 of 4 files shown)


pkgs/_analyzer_macros/bin/server.dart
```diff
@@ -0,0 +1,35 @@
...
+/// Only works for one path, see `packageConfig` path hardcoded below.
+void main(List<String> args) async {
+  final home = Platform.environment['HOME']!;
+  injected.macroImplementation = await AnalyzerMacroImplementation.start(
+      protocol: Protocol(
+          encoding: ProtocolEncoding.binary, version: ProtocolVersion.macros1),
+      // TODO(davidmorgan): this needs to come from the analyzer, not be
+      // hardcoded.
+      packageConfig:
+          Uri.file('$home/git/macros/.dart_tool/package_config.json'));
```
jakemac53, line 31: I believe this can be `Isolate.packageConfig` for now? At least it won't be specific to your file system layout then 😄

jakemac53: Actually nvm I think there won't be a package config when this actually runs.

However, maybe a `-D` variable, then the path gets compiled into the script as a constant.

jakemac53 (approved):

pkgs/_analyzer_macros/bin/server.dart
```diff
@@ -0,0 +1,35 @@
...
+///
+/// dart compile kernel bin/server.dart
+/// cp bin/server.dill ~/opt/dart-sdk-be/bin/snapshots/analysis_server.dart.snapshot
+///
+/// Then restart, VSCode: Ctrl+Shift+P, Restart Analysis Server.
+///
+/// Only works for one path, see `packageConfig` path hardcoded below.
+void main(List<String> args) async {
+  final home = Platform.environment['HOME']!;
+  injected.macroImplementation = await AnalyzerMacroImplementation.start(
```
jakemac53, line 25: I am curious how we are going to handle enabling/disabling this in general?

If we are going to attempt to support old and new macros together, then the way we are injecting things right now won't really work, because it is a global setting, and you might have multiple projects open with different macro implementations.

pkgs/_analyzer_macros/bin/server.dart
davidmorgan: Right.

`-D` sounds good; done and updated the example command.

davidmorgan (commented):
Thanks, PTAL.

jakemac53 (approved):

pkgs/_analyzer_macros/bin/server.dart
davidmorgan: I'm sure I wrote a reply to this, not sure where it went...

My expectation is that we'll switch from one to the other without supporting both, and possibly with some downtime in between when we support neither.

It would be more work+complexity to support both for external users, e.g. we'd have to do something about `package:macros` being used in two different ways, or use two different packages.

For our own use for development I guess one or the other is enough, or we can add something more fine-grained if needed.

jakemac53: An atomic switch will probably be a bit tricky given all the tests which work with the existing API as well as the JSON macro (which lives in the SDK but we could move it out which might help).

It might still be the easiest answer though.

PR #121, opened 2024-10-25
Add support for definitions implementing other definitions

jakemac53:
Towards https://github.com/dart-lang/macros/issues/103

Adds an `implements` list to `Definition.clazz`, which:

- Replaces the default `implements Object` and instead lists all the implemented types (which transitively gives `Object`).
- Adds all inherited properties and associated code to the constructors and schemas
- Does not re-generate getters, these are inherited via `Implements`

I also added a base type `Declaration` because it is both useful and also serves as an example usage.

I think one good follow-up would be to add `is<Type>` and `as<Type>` getters to anything which is implemented by another type, for each of the types which implement it. We at least need some mechanism for down casting extension types to their known subtypes. One tricky part with that though is ideally we don't want to inherit those methods in the subtypes, because they won't all make sense?

### DIFF
--- a/pkgs/dart_model/lib/src/dart_model.g.dart
+++ b/pkgs/dart_model/lib/src/dart_model.g.dart
@@ -137,29 +137,50 @@ extension type MetadataAnnotation.fromJson(Map<String, Object?> node)
   QualifiedName get type => node['type'] as QualifiedName;
 }
 
-/// An interface.
-extension type Interface.fromJson(Map<String, Object?> node) implements Object {
+/// Base type for all declarations
+extension type Declaration.fromJson(Map<String, Object?> node)
+    implements Object {
   static final TypedMapSchema _schema = TypedMapSchema({
     'metadataAnnotations': Type.closedListPointer,
-    'members': Type.growableMapPointer,
-    'thisType': Type.typedMapPointer,
     'properties': Type.typedMapPointer,
   });
-  Interface({
+  Declaration({
     List<MetadataAnnotation>? metadataAnnotations,
-    NamedTypeDesc? thisType,
     Properties? properties,
   }) : this.fromJson(Scope.createMap(
           _schema,
           metadataAnnotations,
-          Scope.createGrowableMap(),
-          thisType,
           properties,
         ));
 
-  /// The metadata annotations attached to this interface.
+  /// The metadata annotations attached to this declaration.
   List<MetadataAnnotation> get metadataAnnotations =>
       (node['metadataAnnotations'] as List).cast();
+
+  /// The properties of this declaration.
+  Properties get properties => node['properties'] as Properties;
+}
+
+/// An interface.
+extension type Interface.fromJson(Map<String, Object?> node)
+    implements Declaration {
+  static final TypedMapSchema _schema = TypedMapSchema({
+    'members': Type.growableMapPointer,
+    'thisType': Type.typedMapPointer,
+    'metadataAnnotations': Type.closedListPointer,
+    'properties': Type.typedMapPointer,
+  });
+  Interface({
+    NamedTypeDesc? thisType,
+    List<MetadataAnnotation>? metadataAnnotations,
+    Properties? properties,
+  }) : this.fromJson(Scope.createMap(
+          _schema,
+          Scope.createGrowableMap(),
+          thisType,
+          metadataAnnotations,
+          properties,
+        ));
 
   /// Map of members by name.
   Map<String, Member> get members =>
@@ -167,9 +188,6 @@ extension type Interface.fromJson(Map<String, Object?> node) implements Object {
 
   /// The type of the expression `this` when used in this interface.
   NamedTypeDesc get thisType => node['thisType'] as NamedTypeDesc;
-
-  /// The properties of this interface.
-  Properties get properties => node['properties'] as Properties;
 }
 
 /// Library.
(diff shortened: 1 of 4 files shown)


jakemac53:
For the follow up, one idea is to add extensions on the implemented types, when generating the type which implements them. That way you don't actually need knowledge of classes implement what types when generating the code for the supertypes, you can locally add the getters when generating each type.

davidmorgan:
Thanks Jake! Fun stuff. I thought about it for a while today, and have one suggestion to throw into the pile :)

How far do we get if we weaken `Declaration` so it's more like a pure interface than a base class?

 - `Declaration` keeps the getters, loses the schema and no-name constructor, does not appear in the JSON schema:
```
/// Base type for all declarations
extension type Declaration.fromJson(Map<String, Object?> node)
    implements Object {
  /// The metadata annotations attached to this declaration.
  List<MetadataAnnotation> get metadataAnnotations =>
      (node['metadataAnnotations'] as List).cast();

  /// The properties of this declaration.
  Properties get properties => node['properties'] as Properties;
}
```
 - `Interface` and `Member` continue to `implement Declaration` so they can be passed to a method that wants to operate on a `Declaration`
 - This lets you pass `Interface` or `Member` to a method that wants a `Declaration`

Then "code that wants to downcast" we solve as a different problem, two options come to mind:

 - The union types that we already have, e.g. we could have a new union type `TopLevelEntry`. If a `TopLevelEntry` is always going to hold only values that implement `Declaration` then this looks bad ... but I kinda like that this way you could add things to `TopLevelEntry` that are _not_ forced to have the same two `Declaration` properties. We could even generate `isDeclaration` and `asDeclaration` in the union type.
 - Or, general downcasts with a `type` field. I'm a bit hesitant about adding this everywhere, it is a more powerful/flexible feature than union types, I'd like to have the data model be as weak as possible while still doing what we need :)

What do you think? :)

jakemac53:
I would be fine with essentially not encoding the type hierarchy as a part of the schema itself, which is I think the point of what you are suggesting here?

Then, even if we do have a type hierarchy in the Dart code, it still isn't breaking for the protocol to for instance add non-declaration things to the top level members, because its just a flat list of possible types and you can't do anything until checking the type essentially.

> * Or, general downcasts with a `type` field. I'm a bit hesitant about adding this everywhere, it is a more powerful/flexible feature than union types, I'd like to have the data model be as weak as possible while still doing what we need :)

The reason I think that we need this, is we don't currently have any sort of reasonable way at all to check what type something is, once you have a hold of an "instance". You can only check the type if you are looking at the union type entry, but then all info about it is basically lost. This makes it not possible to write a function which takes in a `FunctionDeclaration`, and then checks internally if it is actually a `MethodDeclaration` to do some slight modification of its logic (we have code which does this today fwiw, it isn't contrived). We could just not attempt to have any sort of type hierarchy at all, but I think it is very useful to have.

davidmorgan:
> I would be fine with essentially not encoding the type hierarchy as a part of the schema itself, which is I think the point of what you are suggesting here?
> 
> Then, even if we do have a type hierarchy in the Dart code, it still isn't breaking for the protocol to for instance add non-declaration things to the top level members, because its just a flat list of possible types and you can't do anything until checking the type essentially.

Right, that seems better if it does what we need.

> > * Or, general downcasts with a `type` field. I'm a bit hesitant about adding this everywhere, it is a more powerful/flexible feature than union types, I'd like to have the data model be as weak as possible while still doing what we need :)
> 
> The reason I think that we need this, is we don't currently have any sort of reasonable way at all to check what type something is, once you have a hold of an "instance". You can only check the type if you are looking at the union type entry, but then all info about it is basically lost. This makes it not possible to write a function which takes in a `FunctionDeclaration`, and then checks internally if it is actually a `MethodDeclaration` to do some slight modification of its logic (we have code which does this today fwiw, it isn't contrived). We could just not attempt to have any sort of type hierarchy at all, but I think it is very useful to have.

Great, thanks, it's much easier to work with specific examples :)

We do have enough information on the wire to figure out the type today, as long as the instance is in the `Model`, from the path. The instance must either be from somewhere with the exact type declared--we'd need to generate this data--or from a union type, in which case we can read the type from that. We could make "having a getter for type" optional per type and generate an enum of all the types that opt in for it to return.

That gives you a flat list of types, not a type hierarchy, which I think may be "correct": an old client has no way to know that a new type that was added is supposed to be a subtype of an existing type, it will be treated as an unknown type. So it would be more like `.runtimeType == MethodDeclaration` than `is MethodDeclaration`.

Once we can talk about type we _could_ send information per type, a bit like having static fields on the types, so for example you could describe the properties of every `MethodDeclaration` without having to add a field on every instance that always has to carry the same value. So `FunctionDeclaration` could have a "static" `isMethodDeclaration` or `declarationType` that delegates to the per-type data for every implementing type.

If the above sounds unnecessarily complicated ;) it's because I'm trying to reconcile the simplicity of union types with the simplicity (from an entirely different point of view) of subtypes, maybe some of that is helpful, maybe not ;)

I guess, let's discuss tomorrow? I'll anyway give it some more thought and see if any other ideas arrive :) ... thanks.

jakemac53:
I do think it is worth chatting tomorrow about the specifics of how we want to move forward here. I pushed a change to this which removes Declaration from the schema (and adds support for removing types from the schema, as well as removing their constructors if they really just pure interfaces).

I believe that can land without us answering the larger questions here as it isn't actually directly related? Unless we don't want things to actually `implement` the types they "implement", and just copy all the members instead. But, then we lose the ability to for instance pass a `MethodDeclaration` to something that wants a `FunctionDeclaration`, without an explicit "cast". It could be reasonable though to just force going through a function that does the upcast, so we can discuss that as well.

jakemac53:
Another random thought here is that if we do support upcasting unconditionally via a getter, that doesn't improve the situation (if for some reason a method is no longer always a function, its breaking because that upcast is now no longer unconditionally valid).

So, I think we would always want to pair it with an `bool get is*` getter.

davidmorgan (approved):

davidmorgan:
Thanks Jake! Yes, I think this is now good to merge, leaving some fun open questions for discussion :)

I experimented a bit on top of this PR, trying two different directions:

1. Support checking the wire type, and casting, [PR](https://github.com/dart-lang/macros/pull/123/commits/3a923d27c300a6ff615e88221482224fd697c0bf#diff-4fd55a69b704425884152a8d86e2c561e6c354ac91a1091d80ed6489bbf0d776R227)
2. Add `declarationType` property and `DeclarationType` enum [PR](https://github.com/dart-lang/macros/pull/122/commits/b5b93a330a6d22355878bb50a5b8f732b8f28b05#diff-4fd55a69b704425884152a8d86e2c561e6c354ac91a1091d80ed6489bbf0d776R229)

Implementation notes from 1: getting the wire type from the path works, although of course I just hacked it rather than generating properly. We could push all the `is`/`as` methods onto a `Cast` type to avoid polluting all the interfaces quite so much. The `wireType` getter I put in a few places to show different possibilities, presumably we would pick one.

Implementation notes from 2: we end up wanting a field with a fixed value, so `declarationType` is always `DeclarationType.member` for a `Member` and always `DeclarationType.interface` for an `Interface`. There should be a nice way to do this by putting the fixed value in the schema, it will get written to the buffer once, I believe this also works as we want if it changes in a later version.

In both cases the nicest way I could find to add a test equivalent to the example you mentioned in the builder (`executor/builder_impls.dart`, `_buildFunctionAugmentation`), was to add "cast or return null" `maybe` methods and use null checking for "promotion", see the `produceOutput` methods in the tests added.

The two approaches end up looking fairly similar for the simple case in the test.

The manually-added enums approach is more flexible but requires you to add the enum and cast helper methods yourself. We could automate that, so that for an interface you automatically get an enum with all implementers and cast methods for them in the interface. (Or in a `Cast` type so it doesn't pollute the namespace so much).

There is a question as to whether deducing the type from the path is a good idea at all; if we avoid using that trick then the two approaches look even more similar :) storing the type as a "static field" in the schema, as I suggested for the manual enums, is extremely natural. I think the main difference that remains is whether we have one enum across all types or if there are multiple smaller enums and corresponding smaller "cast" APIs.

PR #122, opened 2024-10-29
Type hierarchy use enum

davidmorgan:


jakemac53:
I definitely prefer this approach, especially if we can move the type string to the schema as you had mentioned. Probably as a list of interfaces, but we could keep it as just one and generate in the code a list of the interfaces for each type. Either way would be OK with me.

davidmorgan:
Sounds good, happy turn this into a real PR, or let me know if you want to take a shot at it. Thanks.

jakemac53:
Feel free to run with it!

PR #124, opened 2024-10-29
Add benchmark_generator

davidmorgan:
This is mostly old code, there is a bit of unused code around adding random "cache busters" which will be needed for any kind of actual benchmark.

Running like this:

```
dart bin/main.dart large macro 64
```

Generates source under `goldens/foo/lib/generated/large` which can be run with

```
dart ~/git/macros/pkgs/_macro_tool/bin/main.dart \
    --workspace=$HOME/git/macros/goldens/foo \
    --packageConfig=$HOME/git/macros/.dart_tool/package_config.json \
    --script=$HOME/git/macros/goldens/foo/lib/generated/large/a0.dart \
    --host=analyzer --watch
```

which on changing `a0.dart` shows a refresh time of ~9s. Digging through my old spreadsheets I found "9.289s" as the entry for this benchmark with the v1 macro, so it looks like we are equal or even slightly faster :) ... the v2 macro still needs a bit more error checking though.

### DIFF
--- a/tool/benchmark_generator/README.md
+++ b/tool/benchmark_generator/README.md
@@ -1,0 +1,16 @@
+# Benchmark Generator
+
+Generates code that uses macros, for benchmarking.
+
+Example use:
+
+```
+dart bin/main.dart large macro 64
+dart ~/git/macros/pkgs/_macro_tool/bin/main.dart \
+    --workspace=$HOME/git/macros/goldens/foo \
+    --packageConfig=$HOME/git/macros/.dart_tool/package_config.json \
+    --script=$HOME/git/macros/goldens/foo/lib/generated/large/a0.dart \
+    --host=analyzer --watch
+```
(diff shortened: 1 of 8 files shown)


jakemac53 (approved):
Can you add a README.md with the instructions in the top comment please :)

tool/benchmark_generator/README.md
```diff
@@ -0,0 +1,16 @@
+# Benchmark Generator
+
+Generates code that uses macros, for benchmarking.
+
+Example use:
+
+```
+dart bin/main.dart large macro 64
+dart ~/git/macros/pkgs/_macro_tool/bin/main.dart \
```
jakemac53, line 9: It would be nice if this was something anybody could copy/paste and run from a specified path in the repo (in other words, if it was all relative paths)

PR #126, opened 2024-10-30
Return macro metadata from analyzer.

davidmorgan:
Wire up macro metadata in queries in the analyzer implementation.

Use `parseAnnotation` in the analyzer codebase and the converter in `_analyzer_cfe_macros`.

`dart_model` type `MetadataAnnotation` holds an `Expression` instead of a `QualifiedName`.

Add goldens/foo/lib/metadata.dart and goldens/foo/lib/metadata.analyzer.json which shows it working for a couple of primitive named fields.

`dart_model` definitions changed to add `createInBuffer: true` so they can be added to query results.

### DIFF
--- a/goldens/foo/lib/metadata.analyzer.json
+++ b/goldens/foo/lib/metadata.analyzer.json
@@ -1,0 +1,164 @@
+{
+  "uris": {
+    "package:foo/metadata.dart": {
+      "scopes": {
+        "Foo": {
+          "members": {
+            "": {
+              "requiredPositionalParameters": [],
+              "optionalPositionalParameters": [],
+              "namedParameters": [],
+              "properties": {
+                "isAbstract": false,
+                "isConstructor": true,
+                "isGetter": false,
+                "isField": false,
+                "isMethod": false,
+                "isStatic": false
+              }
+            }
+          },
+          "metadataAnnotations": [
+            {
+              "expression": {
+                "type": {
+                  "reference": {
+                    "type": "ClassReference",
+                    "value": {}
+                  },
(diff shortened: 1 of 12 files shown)


goldens/foo/lib/metadata.analyzer.json
```diff
@@ -0,0 +1,164 @@
...
+              }
+            }
+          },
+          "metadataAnnotations": [
+            {
+              "expression": {
+                "type": {
+                  "reference": {
+                    "type": "ClassReference",
+                    "value": {}
```
johnniwinther, line 27: How much information do you need for the references to have meaning? A name, uri, enclosing declaration name, ... ?

johnniwinther (approved):

goldens/foo/lib/metadata.analyzer.json
davidmorgan: The first obvious thing is that a macro needs to be able to identify its own annotation(s), which would "mostly work" with just name but needs name+uri to be correct.

I guess the two hardest use cases we have are the dependency injection framework example of aggregating "providers", and the general case of wanting to be able to output expressions into macro-generated code.

johnniwinther: I'll add these the `Reference`s.

jakemac53 (approved):

davidmorgan:
Looks like this is still incorrect when there are lists containing union types, e.g. `List<Argument>`, the converter doesn't handle those correctly.

It's easy enough to fix, planning to send a PR with fix and example macro tomorrow.

PR #128, opened 2024-10-31
Add PrimitiveParams macro.

davidmorgan:
Fix converter for lists of union types.


goldens/foo/lib/metadata.analyzer.json
```diff
@@ -55,20 +55,26 @@
...
-                        "type": "IntegerLiteral",
-                        "value": {
-                          "text": "23"
+                      "type": "NamedArgument",
+                      "value": {
+                        "name": "anInt",
+                        "expression": {
+                          "type": "IntegerLiteral",
+                          "value": {
+                            "text": "23"
```
johnniwinther, line 76: I've update the `BooleanLiteral`, `IntegerLiteral` and `DoubleLiteral` to hold their primitive `value` where possible (ints are a bit weird). It might be good to pass that along - in particular since evaluation (coming shortly) will only provide the `value` and not the `text`.

pkgs/_test_macros/lib/literal_params.dart
```diff
@@ -0,0 +1,119 @@
...
+      for (final entry in namedArguments.entries)
+        ' // ${entry.key}: ${entry.value}, ${entry.value.runtimeType}',
+    ].join('\n'))));
+  }
+}
+
+// TODO(davidmorgan): common code for this in `dart_model` so macros don't
+// all have to write expression evaluation code.
+extension ExpressionExtension on Expression {
+  Object get evaluate => switch (type) {
```
johnniwinther, line 78: This is done more extensively in https://dart-review.googlesource.com/c/sdk/+/392440

johnniwinther (approved):

pkgs/_test_macros/lib/literal_params.dart
```diff
@@ -0,0 +1,119 @@
...
+      annotation: QualifiedName(
+          uri: 'package:_test_macros/literal_params.dart',
+          name: 'LiteralParams'),
+      runsInPhases: [2]);
+
+  @override
+  Future<void> buildDeclarationsForClass(
+      ClassDeclarationsBuilder builder) async {
+    final annotation = builder
+        .target.metadataAnnotations.first.expression.asConstructorInvocation;
```
jakemac53, line 58: We should probably add a TODO for checking the type of the annotation (in practice you will want to search for a specific one).

jakemac53 (approved):

pkgs/_test_macros/lib/literal_params.dart
davidmorgan: Good point, thanks! Done.

pkgs/_test_macros/lib/literal_params.dart
davidmorgan: Thanks! I expect we need a higher level discussion soon to figure out where all this is headed. Or one of us could throw some thoughts on an issue/doc.

No rush though, going to push on the other tracks (performance, data model) a bit next. Thanks.

davidmorgan (commented):
Thanks!

PR #131, opened 2024-11-01
[Do not submit] Add model digest

jakemac53:
I created this simplified version of a model digest mostly out of curiosity to see if it could actually work, and it turns out it doesn't :). The test fails because the digests are different.

I think that just hashing the raw bytes can never really be relied upon because of all the pointers in the bytes - unless it is built in exactly the same order the bytes will inevitably be different.

There is another issue that if the Model is a not the root of the buffer, it will capture more information than we want, in particular info that is not hermetic (for instance, request IDs).

So, I think we will need a different approach here.


jakemac53:
cc @davidmorgan

PR #132, opened 2024-11-04
Use cached results based on hash of query responses

jakemac53:
Implements caching into the MacroHost class. This is not necessarily the most efficient place to do it, but it is the most general place.

Adds methods on QueryService to start and stop tracking queries. This will need to get more sophisticated later on, to support concurrent macros.

Currently, likely due to the poor performance of the hashing, this is essentially equivalent to no smart invalidation at all, maybe even slower by around 10%. As such we should consider whether or not we want to actually land this in its current state. I would probably lean towards landing it as is though, and then trying to iterate on faster hashing solutions afterwords.

### DIFF
--- a/pkgs/_macro_host/lib/src/macro_cache.dart
+++ b/pkgs/_macro_host/lib/src/macro_cache.dart
@@ -36,6 +36,10 @@ class MacroResultsCache {
     );
   }
 
+  /// Returns a cached result for the given augmentation request, if there is
+  /// one and it is still valid.
+  ///
+  /// Otherwise, invalidates the cache for this request and returns `null`.
   Future<AugmentResponse?> cachedResult(
       QualifiedName annotation, AugmentRequest request) async {
     final cacheKey = (
--- a/pkgs/_macro_host/lib/src/macro_host.dart
+++ b/pkgs/_macro_host/lib/src/macro_host.dart
@@ -69,6 +69,7 @@ class MacroHost {
   /// Sends [request] to the macro triggered by [annotation].
   Future<AugmentResponse> augment(
       QualifiedName annotation, AugmentRequest request) async {
+    // TODO: Save the query results or pre-emptively send them to the macro?
     final cached = await _macroResultsCache.cachedResult(annotation, request);
     if (cached != null) return cached;
 
--- a/pkgs/dart_model/lib/src/dart_model.dart
+++ b/pkgs/dart_model/lib/src/dart_model.dart
@@ -33,7 +33,7 @@ extension ModelExtension on Model {
   /// An identity hash for `this`, used for comparing query results.
   ///
   /// TODO: A faster/better implementation?
-  int get identityHash => const DeepCollectionEquality.unordered().hash(node);
+  int get identityHash => const DeepCollectionEquality().hash(node);
 
   /// Looks up [name] in `this`.
   ///


pkgs/dart_model/lib/src/dart_model.dart
```diff
@@ -28,6 +30,11 @@ extension ParentInterface on Member {
 }
 
 extension ModelExtension on Model {
+  /// An identity hash for `this`, used for comparing query results.
+  ///
+  /// TODO: A faster/better implementation?
+  int get identityHash => const DeepCollectionEquality.unordered().hash(node);
```
jakemac53, line 36: Fwiw, it looks like the implementation of this does not use `entries`, but instead uses `keys` and then reads the value of the key, so this might explain some of the slowness.

jakemac53: Ok, yeah wow I hacked a version of package:collection to use `entries` and it is at least twice as fast, and now faster overall with the cached results than without. I might look at  upstreaming that, and if not then we can have our own implementation.

pkgs/_macro_host/lib/src/macro_host.dart
```diff
@@ -0,0 +1,188 @@
...
+  Future<Set<int>> queryMacroPhases(
+      Uri packageConfig, QualifiedName annotation) async {
+    await _ensureRunning(annotation);
+    return _hostService._macroState[annotation.asString]!.phases;
+  }
+
+  /// Sends [request] to the macro triggered by [annotation].
+  Future<AugmentResponse> augment(
+      QualifiedName annotation, AugmentRequest request) async {
+    final cached = await _macroResultsCache.cachedResult(annotation, request);
```
davidmorgan, line 72: It took me a while to figure out what this is doing: rerun the queries from last time, discard the output from the cache if anything changed.

It looks like it would make sense to also cache the query results here, so that if we do run the macro then the ones we fetched to check invalidation can be used instead of requested again?

I guess a TODO for that would make sense :)

pkgs/_macro_host/lib/src/macro_cache.dart
```diff
@@ -0,0 +1,97 @@
...
+      annotation: annotation.asRecord,
+      target: request.target.asRecord,
+      phase: request.phase,
+    )] = (
+      queries: queryResults.map((q) => q.asCachedResult),
+      response: response
+    );
+  }
+
+  Future<AugmentResponse?> cachedResult(
```
davidmorgan, line 39: It took me a while to figure out what's going here, particularly rerunning of queries :)

It's hard to come up with a fully descriptive name for this method ... rerunQueriesAndReturnCachedResultIfResultsUnchanged? :) ... maybe it would make sense to split up so the rerunning of the queries is separate from the check?

That would also help address my other comment which is that it would be nice not to throw away the query results, presumably they should go into some other cache so the macro can use them.

Feel free to leave TODOs for any/all of these :)

pkgs/dart_model/lib/src/dart_model.dart
davidmorgan: Ordered comparison might be faster, too? If the ordering is sufficiently stable...

jakemac53: yeah about 10%, switching to that.

pkgs/_macro_host/lib/src/macro_host.dart
jakemac53: The queries themselves could change too, but it might make sense to try something like that

pkgs/_macro_host/lib/src/macro_cache.dart
jakemac53: I added a doc comment to at least explain this method :)

davidmorgan (approved):
LG, thanks.

PR #133, opened 2024-11-05
use a single hash for the combined query results

jakemac53:
As discussed in our meeting today, this removes about a full second from the large benchmark example.


davidmorgan (approved):
Nice :)

PR #134, opened 2024-11-06
Use Object.hash to hash raw bytes

jakemac53:
Avoids decoding values during hashing just to re-encode them as bytes to hash, when stored as raw bytes. Ultimately this mostly just applies to Strings, but that is still impactful. We also avoid a lot of type checks, instead doing switches based on the known or encoded types.

For now I am just using Object.hash and Object.hashAll but we could explore alternatives later on as well, although it would mean passing a byte sink around.

This drops the total time for the large benchmark another 10% ish on my machine.

### DIFF
--- a/pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
+++ b/pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
@@ -62,13 +62,12 @@ class JsonBufferBuilder {
   /// If [alreadyDereferenced] is `true`, then for types which are pointers,
   /// [pointer] already points at the top of the object, and should not be
   /// followed before reading the object.
-  int identityHash(int pointer,
-      {Type? type, bool alreadyDereferenced = false}) {
+  int fingerprint(int pointer, {Type? type, bool alreadyDereferenced = false}) {
     if (type == null) {
       type = _readType(pointer);
       pointer += _typeSize;
     }
-    return _identityHash(pointer, type,
+    return _fingerprint(pointer, type,
         alreadyDereferenced: alreadyDereferenced);
   }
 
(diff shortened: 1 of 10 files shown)


pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -52,6 +52,87 @@ class JsonBufferBuilder {
...
+      case Type.boolean:
+        return _buffer[pointer];
+      case Type.anyPointer:
+        return identityHash(pointer);
+      case Type.stringPointer:
+        final length = _readLength(pointer);
+        pointer += _lengthSize;
+        return Object.hashAll(_buffer.sublist(pointer, pointer + length));
+      case Type.closedListPointer:
+        final length = _readLength(pointer);
```
davidmorgan, line 107: It would be good to keep code that knows about the byte layout of these types in their respective part files, they can just be static methods if that's what they need to be.

pkgs/dart_model/lib/src/json_buffer/json_buffer_builder.dart
```diff
@@ -52,6 +52,87 @@ class JsonBufferBuilder {
...
+  ///
+  /// Any nested pointers use the hash of the values they point to.
+  ///
+  /// If [type] is provided, [pointer] should point directly at the object.
+  /// Otherwise a [Type] will be read first, followed by the value.
+  ///
+  /// If [alreadyDereferenced] is `true`, then for types which are pointers,
+  /// [pointer] already points at the top of the object, and should not be
+  /// followed before reading the object.
+  int identityHash(int pointer,
```
davidmorgan, line 65: Maybe "fingerprint"? That would help keep it distinct from `hashCode` and `==`.

davidmorgan (approved):
A couple of high level notes:

 - Needs some test coverage :)
 - Pretty sure `Object.hash` is not what we want in the end, 32 bit Jenkins hash is pretty small; doesn't need changing now though

jakemac53:
> * Needs some test coverage :)

Huh, I definitely wrote a test at some point previously covering this but I can't find it... 🤔 . I think I must have never landed it lol.

jakemac53:
Aaaand now I am on an adventure fixing growable maps 🤣 (apparently if they have growable maps as values they stack overflow 😄  )

jakemac53:
> * Pretty sure `Object.hash` is not what we want in the end, 32 bit Jenkins hash is pretty small; doesn't need changing now though

Yeah it is pretty explicitly documented as "collisions are OK, just might make maps slow" 🤣 . So, we very likely do want to do something else here.

PR #135, opened 2024-11-07
Improvements to macro_tool for benchmarking.

davidmorgan:
Pull out functionality to top-level commands.

Add support for benchmarking: both macro applications and analysis of code as it would be with macro output.

Add lots of examples to `README.md`.


pkgs/_macro_tool/README.md
```diff
@@ -1,15 +1,74 @@
-A tool for running macro examples end to end.
+# macro_tool
```
jakemac53, line 1: nit: in general we don't include these headers, the context should already be obvious

pkgs/_macro_tool/README.md
```diff
@@ -1,15 +1,74 @@
...
+  benchmark_apply: benchmarks applying macros
+  benchmark_analyze: benchmarks analyzing macros
+  bust_caches: modifies files like the benchmarks do
+  watch: loops watching for changes to the specified script and applying
+
+## Examples
+
+All examples are run from the root of this repo.
+
+Benchmarks required that you first create sources to benchmark with:
```
jakemac53, line 21:
```suggestion
Benchmarks require that you first create sources to benchmark with:
```

pkgs/_macro_tool/README.md
```diff
@@ -1,15 +1,74 @@
...
+  --script=goldens/foo/lib/json_codable_test.dart \
+  apply patch_for_cfe run revert
 ```
 
-You can also enable watch mode using `--watch`.
+Watch for changes to a script, applying macros when it changes, writing the
+output to disk and reporting how long it took:
+
+```bash
+dart pkgs/_macro_tool/bin/main.dart \
```
jakemac53, line 59: When running on a small example like this, using the JIT VM, the results are probably very misleading until you do several rounds of edits. It might be worth mentioning here?

pkgs/_macro_tool/lib/analyzer_macro_runner.dart
```diff
@@ -0,0 +1,89 @@
...
+      {required this.workspacePath, required this.packageConfigPath})
+      : sourceFiles = Directory(workspacePath)
+            .listSync(recursive: true)
+            .whereType<File>()
+            .where((f) => f.path.endsWith('.dart'))
+            .map((f) => SourceFile(f.path))
+            .toList() {
+    final contextCollection =
+        AnalysisContextCollection(includedPaths: [workspacePath]);
+    analysisContext = contextCollection.contexts.first;
```
jakemac53, line 39: Just for sanity:

```suggestion
    analysisContext = contextCollection.contexts.single;
```

pkgs/_macro_tool/lib/analyzer_macro_runner.dart
```diff
@@ -0,0 +1,89 @@
...
+      injected_analyzer.macroImplementation = analyzerMacroImplementation;
+    } else {
+      injected_analyzer.macroImplementation = null;
+    }
+
+    final fileResults = <FileResult>[];
+    final stopwatch = Stopwatch()..start();
+    Duration? firstDuration;
+    for (final sourceFile in sourceFiles) {
+      await analysisContext.applyPendingFileChanges();
```
jakemac53, line 63: This can probably get pulled out of the loop?

```suggestion
    await analysisContext.applyPendingFileChanges();
    for (final sourceFile in sourceFiles) {
```

pkgs/_macro_tool/lib/analyzer_macro_runner.dart
```diff
@@ -0,0 +1,89 @@
...
+          .map((e) => e.toString())
+          .toList();
+
+      final augmentationUnits =
+          resolvedLibrary.units.where((u) => u.isMacroPart).toList();
+      final output = augmentationUnits.singleOrNull?.content;
+
+      fileResults.add(
+          FileResult(sourceFile: sourceFile, output: output, errors: errors));
+      if (firstDuration == null) firstDuration = stopwatch.elapsed;
```
jakemac53, line 81:
```suggestion
      firstDuration ??= stopwatch.elapsed;
```

pkgs/_macro_tool/lib/cfe_macro_runner.dart
```diff
@@ -0,0 +1,109 @@
...
+  final String workspacePath;
+  final String packageConfigPath;
+
+  @override
+  final List<SourceFile> sourceFiles;
+
+  CfeMacroImplementation? cfeMacroImplementation;
+
+  CfeMacroRunner({required this.workspacePath, required this.packageConfigPath})
+      : sourceFiles = Directory(workspacePath)
```
jakemac53, line 26: This looks like it could move into the `MacroRunner` class and then you would extend it from the CFE/analyzer.

The workspace path and packageConfigPath could also become concrete fields in that class.

pkgs/_macro_tool/lib/cfe_macro_runner.dart
```diff
@@ -0,0 +1,109 @@
...
+      '--dart-sdk-summary=${_productPlatformDill.uri}',
+      '--output=${outputFile.path}',
+      for (final sourceFile in sourceFiles) '--source=${Uri.file(sourceFile)}',
+      '--packages-file=$packagesUri',
+      // TODO(davidmorgan): this is so we can pull the generated
+      // augmentation source out of incremental compiler state; find a less
+      // hacky way.
+      '--use-incremental-compiler',
+      // For augmentations.
+      '--enable-experiment=macros',
```
jakemac53, line 87: Is this intended to be `augmentations`?
```suggestion
      '--enable-experiment=augmentations',
```

jakemac53: We might also need enhanced-parts

pkgs/_macro_tool/lib/macro_runner.dart
```diff
@@ -0,0 +1,60 @@
...
+  /// Source files in the workspace.
+  List<SourceFile> get sourceFiles;
+
+  /// Runs macros in all files in the workspace, [sourceFiles].
+  ///
+  /// Returns augmentations and errors for each file.
+  ///
+  /// [injectImplementation] controls whether the `dart_model` macro
+  /// implementation is injected.
+  Future<WorkspaceResult> run({bool injectImplementation = true});
```
jakemac53, line 18: Might want to rename `injectImplentation` to `runMacros` or similar. Really it is controlling whether macros are ran at all, and that it does that by injecting an implementation is more of a detail

pkgs/_macro_tool/lib/main.dart
```diff
@@ -0,0 +1,102 @@
...
+  final commands = args.rest;
+
+  final host = HostOption.forString(args['host'] as String?);
+  final workspace = args['workspace'] as String?;
+  final packageConfig = args['packageConfig'] as String?;
+  final script = args['script'] as String?;
+
+  if (host == null ||
+      workspace == null ||
+      packageConfig == null ||
```
jakemac53, line 32: We really shouldn't need this `packageConfig` argument, can we default it to `Isolate.packageConfig`?

jakemac53 (approved):

pkgs/_macro_tool/README.md
```diff
@@ -1,15 +1,74 @@
...
+  watch: loops watching for changes to the specified script and applying
+
+## Examples
+
+All examples are run from the root of this repo.
+
+Benchmarks required that you first create sources to benchmark with:
+
+```bash
+dart tool/benchmark_generator/bin/main.dart large macro 16
```
jakemac53, line 24: This gives much faster re-compilation of the script itself:
```suggestion
dart run benchmark_generator:main large macro 16
```

Same for the examples below

PR #143, opened 2024-11-19
update gitignore, benchmark readme, and macro tool output

jakemac53:
Some small cleanup stuff:

- Add macro tool output files to gitignore.
- Update benchmark readme to use `dart run`.
- log the total time to re-analyze from the macro_tool (I might be misunderstanding what this is trying to report though).


davidmorgan:
Thanks! Yes, it makes sense to output both timings.

davidmorgan (approved):

PR #145, opened 2024-11-21
Add built_value test macro fields.

davidmorgan:
I think this is enough to be interesting for performance questions.

In particular it checks whether field types are `@BuiltValue` classes, which obviously should use a more advanced query to be fast.


pkgs/_test_macros/lib/built_value.dart
```diff
@@ -94,13 +136,108 @@ class BuiltValueBuilderImplementation implements ClassDeclarationsMacro {
...
+    // Check which field types are annotated with `@BuiltValue`, meaning they
+    // should use nested Builders.
+    //
+    // First, find all the types and query for them.
+    // TODO(davidmorgan): there should be a way to do this in one query.
+    final fieldTypes = <String>{};
+    for (final field in fields) {
+      final qualifiedName = field.value.returnType.asNamedTypeDesc.name;
+      if (qualifiedName.uri != 'dart:core') {
+        fieldTypes.add(qualifiedName.asString);
```
jakemac53, line 157: Why are we adding these as strings and then parsing them below?

davidmorgan: Because `QualifiedName` is an extension type we can't dedupe it with a set, and have to go via `String` in order to do the query once per type instead of once per field.

jakemac53: Oh, hmm that is unfortunate. It seems likely this would be a pretty common use case.

Related question... are we in a "Scope.macro" scope here? If so, when you call `QualifiedName.parse`, is that actually allocating an unused map into the JsonBuffer?

davidmorgan: Yes ... it's similarly annoying in the `_resolveNames` method in `macro_implementation.dart`.

And, yes, it ends up in the JsonBuffer. Definitely room for improvement :)

For this specific case though they should just all be fetched by one initial query, i.e. we want a way to say "and metadata of all field types". Or maybe "any metadata called BuiltValue of all field types"...

jakemac53 (approved):

pkgs/_test_macros/lib/built_value.dart
jakemac53: Definitely seems like something to be done. Not sure what exactly though, don't have any ideas at the moment. It isn't something to block this PR on though in any case.

PR #147, opened 2024-11-21
Use an MD5 hash instead of Object.hash for model hashes

jakemac53:
We can easily swap out the actual hash algorithm here now, in one place.

This is a bit slower (unsurprisingly) than using `Object.hash` etc, a few hundred milliseconds in total per iteration, but is a fairer representation of what we actually want.


jakemac53:
There is one failing test here - `0` and `null` generate the same digest. What value do you think we should use for the bytes of null?

jakemac53:
We could include the type as a part of the digest for everything....

davidmorgan (approved):

PR #153, opened 2024-12-09
add benchmarks for normal Dart classes, serialized to JsonBuffers

jakemac53:
Part of my quest to validate any assumptions we have been making about extension types and how performant they are :D

Here are the new results:

```
 SdkMapsJsonWireBenchmark: 1761ms, 7177227 bytes
     SdkMapsBufferWireBenchmark: 1629ms, 5122684 bytes
    SdkMapsBuilderWireBenchmark: 1553ms, 4859868 bytes
      LazyMapsJsonWireBenchmark: 1079ms, 7177227 bytes
    LazyMapsBufferWireBenchmark:  553ms, 5122684 bytes
LazyWrappersBufferWireBenchmark:  432ms, 2111761 bytes
   BuilderMapsJsonWireBenchmark: 1431ms, 7177227 bytes
BuilderMapsBuilderWireBenchmark:  401ms, 2111761 bytes
RegularClassesBufferWireBenchmark:  589ms, 2111761 bytes

     ProcessSdkMapsJsonWireBenchmark:  214ms, hash 23186292
  ProcessLazyMapsBufferWireBenchmark:  282ms, hash 23186292
ProcessLazyWrappersBufferWireBenchmark:  388ms, hash 23186292
ProcessBuilderMapsBuilderWireBenchmark:  381ms, hash 23186292
ProcessRegularClassesBufferWireBenchmark:   17ms, hash 23186292
```

Note how much faster the processing step is.

If we add up the wire benchmark times and the processing benchmark times, regular classes are faster. This is even though the processing benchmark is actually implemented in a favorable way for the extension types (iterates map entries in order). If you look up by keys, it is much, much slower.

### DIFF
--- a/pkgs/dart_model/benchmark/main.dart
+++ b/pkgs/dart_model/benchmark/main.dart
@@ -10,6 +10,8 @@ import 'lazy_maps_buffer_wire_benchmark.dart';
 import 'lazy_maps_json_wire_benchmark.dart';
 import 'lazy_wrappers_buffer_wire_benchmark.dart';
 import 'lazy_wrappers_buffer_wire_benchmark.dart' as wrapped;
+import 'regular_dart_classes.dart';
+import 'regular_dart_classes.dart' as regular;
 import 'sdk_maps_buffer_wire_benchmark.dart';
 import 'sdk_maps_builder_wire_benchmark.dart';
 import 'sdk_maps_json_wire_benchmark.dart';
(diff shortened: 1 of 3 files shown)


jakemac53:
I would also like to separate the "wire" benchmarks out a bunch, specifically I want to have the following:

- Benchmark for creating the data (but not serializing it)
- Benchmark for serializing the already created data
- Benchmark for deserializing the data (from the serialize benchmark)
- Benchmark for processing the data (can stay as is, hashing is I think a good example)

davidmorgan (approved):

